PR #7, opened 2025-02-26
DOC: Add Sphinx docs

medley56:
This PR adds a docs directory and some stubbed out documentation pages that we should complete and publish to ReadTheDocs. This also adds auto-generated API documentation for all modules in the package. This is customizable and we can tune it if necessary. We are using the built-in "classic" Sphinx theme for now but that can also be customized.


medley56:
@maxinelasp Can I get your review on this?

poetry.lock
maxinelasp, line 1: I believe it is generally considered best practice to commit lock files - why did you remove it?

maxinelasp (commented):
Looks good minus the poetry updates - I am not sure why the poetry.lock file was removed

poetry.lock
medley56: Committing lockfiles is great for application development but a nightmare for library development because the users of your library _won't_ have that lockfile. If all your library developers are using the lockfile, you actually end up failing to discover bugs in your dependency specs until you release the library and users (who don't have your lock file) start reporting that your library doesn't work with the specified dependencies.

https://python-poetry.org/docs/basic-usage/#as-a-library-developer

maxinelasp: Ok, sounds reasonable to me!

maxinelasp:
@BStoneLASP @mmaclay do either of you have any insight into why the tests are still failing? I'm not seeing any changes between this and main that make any code changes, and the tests appear to be passing in main as of yesterday. 

I don't want to have this PR blocked by the failing tests because it's blocking further repository updates to the documentation and cleanup efforts and there aren't any code changes that should be causing test failures as far as I can tell. Should we just merge over the requirement?

PR #14, opened 2025-02-27
Add linting and hook management

medley56:
This PR adds pre-commit and ruff as dependencies for automatic linting and pre-commit hooks. 

Pre-commit is a way to centrally manage git hooks that do things like prevent commits to main and check linting status before allowing the commit to proceed (you can also use hooks other than pre-commit but that's the standard usecase). It's configured in .pre-commit-config.yaml. Pre-commit can also be used with the Pre-commit CI Github app to run hooks automatically on push or merge. This serves as a double check in case developers are not using pre-commit locally to check their own work. To use pre-commit locally, run 
```
poetry lock && poetry install && pre-commit install
```

Ruff is an all-in-one linter that supports dozens of linting standards and tools, including pylint, pyflakes, flake8, Black, Bandit, pycodestyle, pyupgrade, and more. It's almost instantaneous and entirely configurable. I've configured it to run pycodestyle only for now and pre-commit is configured to run ruff before every commit (if you've installed pre-commit as suggested above).


medley56:
@maxinelasp Can I get your review on this?

Issue #18, opened 2025-07-18
Remove or Replace pyhdf dependency in test_compute_elevation.py


mmaclay:
From Matt Watwood:
We had install issues with libera_utils due to curryer on a Mac with an intel processor due to the pyhdf package, is this possible to replace with h5py?

In my quick skim it seems that pyhdf is only used in testing? Specifically, I only see it in tests/test_compute/test_compute_elevation.py . If it can't be replaced, could that pyhdf dependency move over to the [tool.poetry.group.dev.dependencies] rather than a general dependency?
Any run of libera_utils cli tools resulted in a leapsecond error (included in a thread here) from curryer and I'd love to get that cleaned up as part of the tier 1 spice tests.
Is this a configuration piece on the libera_utils side or is it in curryer?

Leap second Error
`An exception occurred while locating the leapsecond file. Suppressing:
Traceback (most recent call last):
  File ".../lib/python3.11/site-packages/curryer/spicetime/leapsecond.py", line 205, in _quiet_load
    load()
  File ".../lib/python3.11/site-packages/curryer/spicetime/leapsecond.py", line 120, in load
    default_kernel = find_default_file()
                     ^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/curryer/spicetime/leapsecond.py", line 60, in find_default_file
    raise FileNotFoundError('Unable to find the default leapsecond kernel file. '
FileNotFoundError: Unable to find the default leapsecond kernel file. Searched directory: .../lib/python3.11/site-packages/data/generic
.../lib/python3.11/site-packages/curryer/spicetime/leapsecond.py:208: UserWarning: Unable to find the default leapsecond kernel file. No leapsecond kernel was loaded!
  warnings.warn('Unable to find the default leapsecond kernel file. No leapsecond kernel was loaded!')`

Issue #17, opened 2025-05-16
Remove RIO/GDAL Dependency for geotiff IO and slicing


mmaclay:
RasterIO is currently used for file IO of geotiffs, and Lat Lon slicing. The GDAL dependency has rendered it uninstallable as of May 2025. We should remove this dependency and opt for a more lightweight solution since we are not using the vast majority of GDAL's functionality to justify it as a dependency.

Current Thoughts:

geotiff - https://github.com/KipCrossing/geotiff
A lightweight, no GDAL library for geotiff file IO and CRS metadata capabilities.

affine - already installed as a dependency, used for coordinate conversion in RasterIO. We can directly use affine to convert between Lat Lon and pixel values for implementing the necessary slicing.

PR #19, opened 2025-07-23
Remove RasterIO / GDAL dependency 

mmaclay:
# PR for Issue #17 (and now also #18)

This PR replaces **rasterIO** and **RIOXarray** with the **GeoTiff** and **Affine** libraries instead.  
There are several new helper functions, and some other small changes required to match the original behavior.  

---

### ✅ Test Results
- **179 Passed**
- **1 Failure**
  - `tests/test_tle.py::TLETestCase::test_ctim_tle_readCTIM`
- **1 Ignored**

- **Libera_Utils** (docker tests pointing to `remove_rio` branch of currier)
  - `334 passed, 4 xfailed, 16 warnings`

### Manually Handling Transformations
- Introduced `get_affine_from_geotiff` to extract and store an **Affine transform** from a GeoTiff object, enabling precise geospatial transformations.
- Added `latlon_to_pixel` to convert longitude/latitude coordinates to raster row/column indices using an Affine transform.
- Replaced `rds.rio.resolution()` with the `resolution` attribute directly from **Affine**.

---

### Raster Data I/O
- Replaced `rioxarray.open_rasterio` with **custom GeoTiff handling**, creating an `xarray.DataArray` with explicit specification of coordinates, resolution, and CRS information.  
- This allows more direct control of the GeoTiff and metadata, without carrying RIO’s default metadata.

---

### PYHDF → H5PY
- Initially fixed and kept the **PYHDF** dependency, but replaced it with **H5PY** (smaller footprint).  
- Unsure if DEMs require HDF4 format.  
- Rewrote the DEM Check test to use **H5PY**, GeoTiff, and Affine transformations.  
  -  **Not tested**, since this test was already skipped and broken after the rasterIO removal.

---

### Dependencies
- Removed:  
  - `pyhdf`, `rasterio`, `rioxarray` (and thus `GDAL`)  
- Added:  
  - `h5py`, `geotiff`, `tifffile`, `affine`, `numcodecs`  
- Also successfully removed **PyShp** after testing locally, in Docker, and via GitHub Actions. Others had to stay for now.

---

### DIFF
--- a/curryer/compute/elevation.py
+++ b/curryer/compute/elevation.py
@@ -98,7 +98,7 @@ class Elevation:
             logger.info('Downloading EGM96 geoid data for pyproj transformer.')
             tg.download_grids(verbose=True)
 
-    def latlon_to_pixel(self, lon, lat, transform):
+    def latlon_to_pixel(self, lon: float, lat: float, transform: Affine) -> Tuple[int, int]:
         """Convert lon/lat to raster row/col indices using the given Affine transform.
 
         Parameters
@@ -125,7 +125,7 @@ class Elevation:
         return int(row), int(col)
 
     @track_performance
-    def get_affine_from_geotiff(self, gtiff):
+    def get_affine_from_geotiff(self, gtiff: GeoTiff) -> Affine:
         """Extract an Affine transform from a GeoTiff object's transform matrix.
 
         Parameters
@@ -498,9 +498,29 @@ class Elevation:
         return files
 
     def _slice_file(self, rds: xr.DataArray, min_lon: float, max_lon: float, min_lat: float, max_lat: float,
-                    orthometric=False, fill_val: float = None) -> xr.DataArray:
-        """Subset a file by a min/max lon/lat region.
+                    orthometric: bool = False, fill_val: float = None) -> xr.DataArray:
         """
+        Subset a DEM file by a min/max lon/lat region.
+
+        Parameters
+        ----------
+        rds : xr.DataArray
+            DEM data array with 'x' (longitude) and 'y' (latitude) coordinates.
+        min_lon, max_lon : float
+            Longitude bounds in degrees (\[-180, 180\]).
+        min_lat, max_lat : float
+            Latitude bounds in degrees (\[-90, 90\]).
+        orthometric : bool, optional
+            If True, return orthometric heights (DEM only). If False, return ellipsoid heights (DEM + geoid).
+        fill_val : float, optional
+            Fill value to check for consistency.
+
+        Returns
+        -------
+        xr.DataArray
+            Subsetted DEM or ellipsoid heights for the region.
+        """
+
         # Use xarray's built-in selection instead of rioxarray
         # Handle edge case that longitude wrapped the international dateline
         if rds['x'].min().item() >= max_lon:


pyproject.toml
```diff
@@ -29,9 +29,12 @@ dependencies = [
     "pandas>=1.5",
     "xarray>=2022.11",
     "netcdf4>=1.6",
-    "pyhdf>=0.11",
-    "rioxarray>=0.15",
-    "rasterio==1.3.9",
+    "h5py>=3.0",
+    "geotiff>=0.2.10",
+    "tifffile>=2021.7.4,!=2022.4.28,<2023.7.0",
```
medley56, line 34: XD `tifffile` must be a fun library to use...

medley56 (changes requested):
My only comment is on the inconsistency of docstrings in the code. Super minor and I only reviewed for coding style. I'm assuming that everything functionally works.

pyproject.toml
```diff
@@ -29,9 +29,12 @@ dependencies = [
...
+    "affine>=2.3",
+    "zarr>=2.12.0,<2.13",
```
mwatwood-cu, line 36: Do you use zarr anywhere explicitly? Not that it needs to be removed, just asking

tests/test_compute/test_compute_elevation.py
```diff
@@ -280,25 +280,92 @@ def test_elevation_local_region_dateline(self):
         npt.assert_allclose(ht_calc2b, ht_calc12[2], rtol=1e-13)
 
     # TODO: Delete or move else where. Used to investigate DEM data sources.
-    @unittest.skip
+    @unittest.skip("Requires test data files - enable if needed for DEM source investigation")
```
mwatwood-cu, line 283: What is the plan to test this and ensure it works? Is there a follow-up issue?

mwatwood-cu:
Like Gavin I reviewed predominately for code style, my only real hanging question is about the skipped test and what the longer term plan is to deal with that. Everything else is minor and mostly stylistic

mwatwood-cu (approved):
Like Gavin I reviewed predominately for code style, my only real hanging question is about the skipped test and what the longer term plan is to deal with that. Everything else is minor and mostly stylistic and I don't need to review it again to see the changes implemented so I will mark as approved.

curryer/compute/elevation.py
```diff
@@ -253,7 +300,33 @@ def load_file(self, filepath: Path) -> xr.DataArray:
...
+            x_coords = [transform * (i + 0.5, 0.5) for i in range(width)]
+            y_coords = [transform * (0.5, j + 0.5) for j in range(height)]
+
+            # Extract just the x and y coordinates from the tuples
+            x_coords = [coord[0] for coord in x_coords]
+            y_coords = [coord[1] for coord in y_coords]
+
+            da = xr.DataArray(
+                array,
+                dims=("y", "x"),
```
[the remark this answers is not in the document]
mmaclay: Great question - I think raster data is conventionally organized as [row, column] (starting with 0,0 as the top left of an image), so what we get from geotiff.read() step is indexed that way. So yes, we index it as `da[y, x]`(but not with the nested double braces as you wrote.)

tests/test_compute/test_compute_elevation.py
mmaclay: I want to loop in @BStoneLASP here since he has likely used this test to look at DEMs in the past and I haven't used it. Not sure if it's necessary or not to have this updated to the new file I/O protocol.

medley56 (approved):

tests/test_compute/test_compute_elevation.py
medley56: @BStoneLASP bump

PR #37, opened 2025-10-23
Update test_tle.py to mock the response from an external API

maxinelasp:
Removes the dependency on the external space-track.org API, which was causing issues when running unit tests. This API is used for CTIM and TSIS-2 but is not needed for other missions, and requires a username/password combination to access. 

Although this API is still used in some instances internally, this makes it so that most developers and users should not need that username/password combo. It also ensures the unit tests can run in a fork without the set up secret values for authentication. 

This PR also includes some weird conflicts with other open PRs but I am planning to merge this after, so you can skip most of the duplicate poetry.lock/.gitignore bits as they will get removed once those other PRs are merged. I can remove those pieces if they are confusing for this review.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,6 +68,7 @@ pytest-cov = "*"
 pytest-randomly = "*"
 ruff = "*"
 ipython = "*"
+responses = "^0.25.8"
 
 #[tool.poetry.scripts]
 #curryer = 'curryer.__main__:main'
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -58,6 +58,7 @@ ipython = ">=8.4"
 
 #[tool.poetry.scripts]
 #curryer = 'curryer.__main__:main'
+responses = "^0.25.8"
```
medley56: This got put in a funky place

medley56 (changes requested):
Needs a rebase and a minor tweak to pyproject.toml. Otherwise, perfect!

pyproject.toml
maxinelasp: Oh you're right, I used `poetry add` and didn't check after. I'll scooch it

PR #38, opened 2025-10-23
Add Minimal de430.bsp Fallback

medley56:
This modified de430_earth_sun_moon_20000101_20350101.bsp contains only the bodies for Earth (399), Earth barycenter (3), Moon (301), and Sun (10) from 2000-01-01 to 2035-01-01. This file was created with spkmerge based on the original de430.bsp but is only 2.5MB in size instead of 115MB.

I've also modified the test kernel configs to reference this generic kernel and they pass because the data for the relevant bodies in the relevant time ranges is the same.

The idea is that this is a fallback/testing kernel that doesn't blow up the repo.


medley56:
I can actually trim this down even more but I wanted it to be somewhat useful as a fallback since it has to go in the generic kernel directory. I wasn't sure if I could put it in tests because it's listed under "sds kernels" in the configs, which seems to point to the generic kernels directory. Could it be moved to the mission-specific kernels part of the config instead, allowing us to put it in test data?

PR #55, opened 2025-10-28
Rebased DEM dep changes and simplified leapsec fix

BStoneLASP:
* Rebased https://github.com/lasp/curryer/pull/19 onto latest
* Simplified https://github.com/lasp/curryer/pull/23 to use existing logic to check if leapsecond file is already loaded before loading, removing need for global variable
* Added upper limit for netcdf library to fix oserrors caused by new netcdf version

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "lasp-curryer"
-version = "0.1.4"
+version = "0.1.3"
 description = "LASP SPICE extentions and geospatial data product generation tools."
 authors = [{ name = "Brandon Stone", email = "brandon.h.stone@colorado.edu" }]
 readme = "README.md"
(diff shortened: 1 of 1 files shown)


tests/test_compute/test_compute_elevation.py
```diff
@@ -291,25 +291,94 @@ def test_elevation_local_region_dateline(self):
         npt.assert_allclose(ht_calc2b, ht_calc12[2], rtol=1e-13)
 
     # TODO: Delete or move else where. Used to investigate DEM data sources.
-    @unittest.skip
+    @unittest.skip("Requires test data files - enable if needed for DEM source investigation")
```
maxinelasp: I think this is the source of the codecov complaints, but I'm fine leaving it

maxinelasp (approved):
Looks good, just asked for a few minor updates and then you should be good to merge!

PR #80, opened 2025-11-10
Remove geotiff and clean up dependencies

### LINKED ISSUE #79: Remove GeoTiff and use Tifffile for geoTiff data and metadata
Implementing [GeoTIFF](https://github.com/KipCrossing/geotiff) helped us leave RasterIO and GDAL behind, but is now lagging on support, and causing a dependency headache, and issues on some platforms (M1):

GeoTIFF is the root of the ridiculous `numcodecs` and `tifffile` pinning in our dependencies as well:
```
numcodecs must be <0.12.0 because geotiff 0.2.10 requires zarr 2.x, and zarr 2.x is only compatible with numcodecs <0.12.0
tifffile 2025.x requires zarr 3.x, but geotiff 0.2.10 requires zarr 2.x. Hence we keep tifffile at ~2023.x versions until geotiff supports zarr 3.x.
```

Since Geotiff is largely built on [tifffile](https://github.com/cgohlke/tifffile) (actively supported and widely used) - and we only use it for basic data and metadata reading of geotiffs, we should refactor the two functions that use `geotiff` to just use `tifffile` directly.

elevation.py
test_compute_elevation.py

functions to replace, both in elevation.py:
- geotiff.GeoTiff() #reads the data and metadata into a geotiff object
- get_affine_from_geotiff() #passes the transformation matrix from geotiff obj into Affine format for conversions / transformations.

This will allow us to significantly update our Pyproject.toml dependencies and should be tested on Several platforms and architectures to ensure we are making good choices.

mmaclay:
This PR refactors elevation.py to use tifffile directly instead of geotiff.

This allowed for significant cleanup of the pyproject.toml, as geotiff was stuck in 2023 dependencies -- limiting `numcodecs` `Zarr` and `tifffile` to outdated versions.

We should test this PR with as many different systems and platforms as possible since it changes pyproject.toml.

Tests failed when I un-pinned netcdf4 and h5py... so keeping them pinned to keep this PR in scope.

So far testing on my system (and Matt W's) is great. I am building and testing on an intel machine today (11.17.25).

This PR has been rebased to Maxine's #85 - My thought is it should be merged right after that one.


maxinelasp (approved):
Looks good to me!

mwatwood-cu:
This solved all the dependency issues I was running into!

mwatwood-cu (approved):
I'm not familiar enough to feel like I can review the functionality of the usage of the tiff files, but on the dependency side, this solved everything I was having issues with.

Issue #84, opened 2025-11-13
Move optional depedendencies to one method


maxinelasp:
## Context

Right now the optional/dev dependencies for Curryer are split across the Poetry dev dependencies group (`[tool.poetry.group.dev.dependencies]`) and the PEP/pip "extras" dependency group (`[project.optional-dependencies]`). This means, if you install with Poetry, you can get different dependencies than  installing with pure pip. 

We should combine these optional dependencies into either a Poetry optional dependencies group or a pip "extras" group. Since Poetry can read from the pip extras group and install them with the `--extras` flag, I think we should move them into the PEP/pip "project.optional-dependencies" group. The drawback of this, is that we cannot use the poetry-specific version syntax (i.e. no carat dependencies such as `^1.2.3`). 

## Requirements
- [ ] Update pyproject.toml file to combine the dependencies
- [ ] update .toml file with new readthedocs URL
- [ ] Confirm that readthedocs now builds, with readthedocs app on the PR branch
- [ ] Update any dev documenation to include new install
- [ ] Investigate if we can tell poetry to install extras by default, since we can assume most devs will want most of the extras

maxinelasp:
Bonus benefit of moving to the PEP compliant dependencies, is that I believe we can then use UV as a drop-in replacement for Poetry.

PR #85, opened 2025-11-13
Updating pyproject.toml and documentation handle extra dependencies

maxinelasp:
See #84 for full context for this change.

I was not able to get numcodecs installed... v0.11.0, which is what is in main, does not provide wheels for Mac ARM 64. I moved it to v0.12.0, which does have those wheels, but I am concerned about what caused that version constraint in the first place - @mmaclay thoughts? Other things I should test against before merging? 

Not included in this PR is that I added the app for readthedocs and checked the box to build off PRs, so hopefully we will see those docs pop up.


maxinelasp:
https://lasp-curryer--85.org.readthedocs.build/en/85/

mmaclay:
> See #84 for full context for this change.
> 
> I was not able to get numcodecs installed... v0.11.0, which is what is in main, does not provide wheels for Mac ARM 64. I moved it to v0.12.0, which does have those wheels, but I am concerned about what caused that version constraint in the first place - @mmaclay thoughts? Other things I should test against before merging?
> 
> Not included in this PR is that I added the app for readthedocs and checked the box to build off PRs, so hopefully we will see those docs pop up.

Yeah fully warranted - the other PR I put in removes that constraint as well. Long story short - we should remove geotiff as a dependency (which is what that PR does). I will bump up the priority of testing that!

maxinelasp:
@mmaclay Great! Glad to hear that, at least. I don't think you need to rush that (I know you're busy with the GeoSIT!). Is it possible/reasonable to use this bump to 0.12.0 temporarily, or should I downgrade to 0.11.0 and open a bug report, with the expectation that your PR will fix it?

mmaclay:
@maxinelasp I built and and tested this fresh environment locally, as my biggest concern would be that the pyproject.toml changes will break geotiff which is currently used in our "--run-extra" tests only which need to be explicitly tested.

All tests passed with a few extra warnings that I'm not too concerned about.

This is definitely related to my imminent PR that will remove `numcodecs` `geotiff` and other limitations in our dependencies, but I think this makes sense to merge first, and then I'll rebase and we can merge the other one soon-after.


A couple questions about the new pyproject.toml structure: other than installing via `poetry install --all-extras` now as a dev, will this change how the repo is installed by libraries that use curryer as a dependency? I.e. CSDS & Libera-utils, does changing that structure affect what is installed for those libraries use-cases?

If so, we should probably test a bit more there (especially since I want to get through geosit before risking breaking curryer even temporarily).

mmaclay:
Tested this on an Intel Mac - good to go.

mmaclay (approved):

PR #89, opened 2025-11-21
Updating instrument intersect ellipsoid for use with high pixel count instrument

mwatwood-cu:
Cleaned up and expanded functionality of the spatial compute function to calculate an intersection with the Earth's ellipsoid.

Expanded the ability to pass in custom pointing vectors without the need to define them in an instrument kernel text format. Specific use case is Libera camera which has ~4 Million pixels and a text based instrument kernel defining them is ~400 MB compared to a ~300 KB netcdf file with the same information.

Naming of functions and parameters along with function structure were updated for clarity and testability.

Testing was expanded and rearranged for clarity. Mocked unit tests were introduced to help test functional usage vs data based results. Integration tests were named as such to use data as an example.

### DIFF
--- a/curryer/compute/spatial.py
+++ b/curryer/compute/spatial.py
@@ -20,11 +20,13 @@ Terms:
 
 import logging
 import time
+import warnings
 
 import numpy as np
 import pandas as pd
 import xarray as xr
 from spiceypy.utils.exceptions import NotFoundError, SpiceyError
+from typing_extensions import deprecated
 
 from .. import spicetime, spicierpy
 from . import constants, elevation
@@ -58,7 +60,7 @@ class SpatialQueries:
             err_value=(np.full((3, 3), np.nan), np.full((3,), np.nan)),
             err_flag=SQF.from_spice_error,
             pass_flag=SQF.GOOD,
-            disable=False
+            disable=False,
         )(_query_rotation_and_position_raw)
     )
 
@@ -74,6 +76,29 @@ class SpatialQueries:
     ):
         """
         Query SPICE for rotation and position, optionally suppressing errors.
+
+        Parameters
+        ----------
+        sample_et : float
+            Ephemeris time to query.
+        instrument : spicierpy.obj.Body
+            Instrument to query.
+        perspective_correction : str
+            SPICE perspective correction to use.
+        observer_id : int, optional
+            Observer NAIF ID (default is Earth).
+        allow_nans : bool, optional
+            If True, suppress SPICE errors and return NaNs instead.
+            Default is False.
+        fixed_frame_name : str, optional
+            Fixed reference frame name (default is ECEF/ITRF93).
+
+        Returns
+        -------
+        tuple
+            Rotation matrix (3x3) and position vector (3,) in fixed frame.
+        SpatialQualityFlags
+            Quality flag indicating results.
         """
         # Dispatch to the appropriate method based on the flag
         if allow_nans:
@@ -137,6 +162,7 @@ def get_instrument_kernel_pointing_vectors(instrument: Union[int, str, spicierpy
     return count, vectors
 
 
+@deprecated("Use get_instrument_kernel_pointing_vectors instead.")
 def pixel_vectors(instrument: Union[int, str, spicierpy.obj.Body]) -> tuple[int, np.ndarray]:
     """Load the pixel or boresight vector(s) for a given instrument.
 
@@ -164,10 +190,9 @@ def pixel_vectors(instrument: Union[int, str, spicierpy.obj.Body]) -> tuple[int,
     """
     # Print deprecation warning and call the new function.
 
-    logger.warning(
+    warnings.warn(
         "`pixel_vectors` is deprecated and will be removed in a future release. "
-        "Please use `get_instrument_kernel_pointing_vectors` instead.",
-        stacklevel=2,
+        "Please use `get_instrument_kernel_pointing_vectors` instead."
     )
     return get_instrument_kernel_pointing_vectors(instrument)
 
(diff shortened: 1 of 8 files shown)


curryer/compute/spatial.py
```diff
@@ -76,13 +130,49 @@ def pixel_vectors(instrument: Union[int, str, spicierpy.obj.Body]) -> tuple[int,
...
     except NotFoundError:
+        logger.debug("Pixel vector data not found. Using the instrument boresight.")
         count = 1
         vectors = np.array([boresight_vector])
 
     logger.debug("Read [%d] pixel vectors for instrument [%s]", count, instrument)
     return count, vectors
 
 
+def pixel_vectors(instrument: Union[int, str, spicierpy.obj.Body]) -> tuple[int, np.ndarray]:
```
maxinelasp: There is an [@deprecated decorator](https://peps.python.org/pep-0702/) you can add to tell the type system this is deprecated, as well.

tests/test_compute/test_compute_spatial.py
```diff
@@ -157,68 +209,64 @@ def test_ellipsoid_intersect_simple(self):
...
+        exp_lla = spicierpy.recgeo(
+            rectan=[xx, yy, zz],
+            as_deg=True,
+            re=constants.WGS84_SEMI_MAJOR_AXIS_KM,
+            f=constants.WGS84_INVERSE_FLATTENING,
+        )
 
-            npix, qry_vectors = spatial.pixel_vectors("CPRS_HYSICS")
+        # ... (Internal Math Verification Logic) ...
+        # Note: I'm trusting your math logic here was correct in the original file
```
maxinelasp: Who is "your" here?

maxinelasp (changes requested):
Not sure any of my comments need code changes, but I want to have a little discussion before merging

curryer/compute/spatial.py
```diff
@@ -209,6 +308,120 @@ def _query(sample_et):
...
+    # Optimization: Ensure contiguous memory for faster matrix multiplication inside the loop
+    pix_vectors = np.ascontiguousarray(pix_vectors)
+
+    logger.debug("Calculating [%s x %s] Earth ellipsoid intercepts for [%s]", pix_count, len(et_times), instrument)
+
+    surface_points = np.full((et_times.size * pix_count, 3), np.nan)
+    spacecraft_positions = np.full((et_times.size * pix_count, 3), np.nan)
+    quality_flags = np.zeros((et_times.size * pix_count,), dtype=np.int64)
+
+    earth_id = spicierpy.obj.Body("EARTH").id
```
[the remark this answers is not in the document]
mmaclay: I suspect we do this a lot in curryer and don't currently have caching set up, but I would like to talk about this and understand it better as it's come up a number of times now.

tests/test_compute/test_compute_spatial.py
```diff
@@ -20,27 +21,97 @@
 
 
 class SpatialTestCase(unittest.TestCase):
-    # Note that many of the routines in `spatial` were originally located in
-    # the `pointing` module, where their original tests might remain.
-
-    def setUp(self) -> None:
+    @classmethod
```
mmaclay: I like using a class method here

mmaclay (changes requested):
I think we should talk about the deprecation and making some more changes so curryer is consistent internally at least. Functionality is all good to go though.

curryer/compute/spatial.py
```diff
@@ -76,13 +130,49 @@ def pixel_vectors(instrument: Union[int, str, spicierpy.obj.Body]) -> tuple[int,
...
+        2D array of boresight (1x3) or pixel vectors (1+,3).
+
+    """
+    # Print deprecation warning and call the new function.
+
+    logger.warning(
+        "`pixel_vectors` is deprecated and will be removed in a future release. "
+        "Please use `get_instrument_kernel_pointing_vectors` instead.",
+        stacklevel=2,
+    )
```
medley56: I'd suggest issuing a DeprecationWarning here with instructions for how to update user code to the new function.
https://docs.python.org/3/library/exceptions.html#DeprecationWarning

curryer/compute/spatial.py
```diff
@@ -209,6 +299,157 @@ def _query(sample_et):
     return pnt_data, sc_data, qf_data
 
 
+def compute_ellipsoid_intersection(
```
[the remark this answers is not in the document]
maxinelasp: Per our discussion yesterday, I'm resolving this comment and we can leave the name.

maxinelasp (approved):
Addressed my concerns, thanks!

curryer/compute/spatial.py
```diff
@@ -20,12 +20,14 @@
...
 import logging
 import time
+import warnings
 from typing import Union
 
 import numpy as np
 import pandas as pd
 import xarray as xr
 from spiceypy.utils.exceptions import NotFoundError, SpiceyError
+from typing_extensions import deprecated
```
mmaclay: This is an external library [typing-extensions](https://pypi.org/project/typing-extensions/) that allows for newer typing features to work in python 310. We would need to add this to the pyproject.toml right? Weirdly I see this library show up in some poetry.lock files locally... but I think we'd need to add it to the toml for consistency.

mmaclay (changes requested):
Thanks for updating all of the tests and internal uses - this looks great and is a big improvement. I rebased and updated this branch locally for my testing and tests are all good to go, warnings expected, etc.

The only things left on my end are:

- [x] 1. please rebase (accept 'yours' for the conflict), run ruff to then remove the Union / Optional calls
- [x] 2. please add typing-extensions to the pyproject.toml

I will then make a fresh environment from the poetry.lock and make sure we are all set.

Looks great!

mmaclay (approved):

Issue #15, opened 2025-02-27
Add pre-commit CI app


medley56:
Pre-commit hooks have been added in #14 but the linting itself is done in the tests actions workflow. Pre-commit CI will basically do this for us outside of an action and report the results.

medley56:
Waiting on merge of #32

PR #90, opened 2025-11-24
Add pre-commit app settings

maxinelasp:
Addresses #15  by adding pre-commit settings for pull requests.

Doesn't look like the code cov validator is supported in the pre-commit.ci. I don't think this is critical to check in CI, so I just skipped it.


mmaclay (approved):

medley56 (approved):
I'm assuming we are planning to improve codecov in a different ticket.

PR #98, opened 2025-12-10
Invert Monte Carlo loop structure & use vectorized operations for performance optimization

### LINKED ISSUE #69: Invert Loop for Monte Carlo Correction
In Monte Carlo, each parameter set is applied to all of the GCP image pairs, so the loop is currently set up as:

```
for parameter_set in parameter_sets:
     for gcp_pair in gcp_pairs:
            ...
```

In production, we should minimize file I/O processes by loading each GCP pair once, geolocating it for each parameter set.
This change will be high risk, affecting checkpoint file writing, and order of operations within the MonteCarlo Loop.

mmaclay:
This PR improves the performance of the Monte Carlo / GCS - current estimates suggest this is going to be really slow in production with significant iterations, so this is required.

Key Changes:

1. Invert Loop in Monte Carlo: reduces file I/O so each image pair is only accessed once instead of n_iterations times.
I tested this thoroughly and ensured it was identical in output to the original version (which I initially kept as deprecated). Once I ensured this, and considering this isn't used yet, I decided to just do a hot swap of the loop structure, removing the original legacy loop structure which does not have any benefits. The new structure should be easier to parallelize.

2. I made small changes to image_match.py and psf.py: implementing vectorization with numpy to replace nested for loops. This yielded a 30% speed increase for the image matching tests (biggest bottle neck for speed).

3. I updated docstrings and added typehints for functions I was modifying already in this PR.


curryer/correction/monte_carlo.py
```diff
@@ -1667,14 +1667,289 @@ def _build_netcdf_structure(config: MonteCarloConfig, n_param_sets: int, n_gcp_p
...
+    return los_vectors_cached, optical_psfs_cached
+
+
+def _load_image_pair_data(tlm_key, sci_key, config, telemetry_loader, science_loader):
+    """
+    Load telemetry and science data for an image pair.
+
+    This function loads and validates both telemetry (L1) and science (L1A)
+    datasets for a single GCP pair. In the current implementation, this is
+    called once per image. In the previous implementation, it was called N
```
maxinelasp: Nice improvement!

curryer/correction/monte_carlo.py
```diff
@@ -1667,14 +1667,289 @@ def _build_netcdf_structure(config: MonteCarloConfig, n_param_sets: int, n_gcp_p
...
+    Create parameter-specific SPICE kernels and apply time offsets.
+
+    This function applies parameter variations by creating modified kernels
+    (CONSTANT_KERNEL, OFFSET_KERNEL) or modifying time tags (OFFSET_TIME).
+    Each parameter set produces different kernels and/or time modifications.
+
+    Args:
+        params: List of (ParameterConfig, parameter_value) tuples
+        work_dir: Path to working directory for kernel files
+        tlm_dataset: pd.DataFrame with spacecraft state
```
maxinelasp: Can you go into more detail on the structure of the DataFrame you're requiring here?If it's an output from another function, what columns are required, etc

curryer/correction/monte_carlo.py
maxinelasp: I know it's inconsistent in this codebase, but could you add typehints to these new functions? I think this kind of code really benefits from them.

maxinelasp (approved):
Nice improvement! This should also help with various efforts to pull out generic structures.

maxinelasp (approved):
Nice documentation improvements!

curryer/correction/monte_carlo.py
```diff
@@ -1667,422 +1726,631 @@ def _build_netcdf_structure(config: MonteCarloConfig, n_param_sets: int, n_gcp_p
...
+        # Collect all image matching results for this parameter set
+        param_image_matching_results = []
+        for pair_idx in range(n_gcp_pairs):
+            result = results_dict.get((param_idx, pair_idx))
+            if result:
+                param_image_matching_results.append(result["image_matching"])
+
+        # Compute aggregate statistics
+        aggregate_stats = call_error_stats_module(param_image_matching_results, monte_carlo_config=config)
+        aggregate_error_metrics = _extract_error_metrics(aggregate_stats)
```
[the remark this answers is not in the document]
mmaclay: Will be addressed in [issue 100](https://github.com/lasp/curryer/issues/100)

Issue #52, opened 2025-10-27
Drop Support for Python 3.9


medley56:
# Summary

Python 3.9 is EOL Oct 10, 2025. We can drop support for it now.

https://devguide.python.org/versions/

maxinelasp:
Closed by #101

Issue #86, opened 2025-11-18
Fix Deprecated `datetime` Calls to `utcnow`


medley56:
# Context

Datetime has deprecated the `utcnow` function in datetime, with a preference to using the more explicit API `datetime.now(datetime.UTC)`, which is compatible with any timezone spec.

We're getting warnings during usage like the following:

```
  /opt/venv/lib/python3.12/site-packages/curryer/kernels/writer.py:175: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    created=datetime.utcnow().strftime("%Y-%m-%d"),
```

# Requirement

- Replace usages of `datetime.utcnow()` with `datetime.now(UTC)`

# Considerations

There are likely many places in the codebase where this API is used and datetime is going to deprecate it soon so we should purge the package of usages.

mmaclay:
Note that pandas has a similar call that is not deprecated yet:

`t0 = pd.Timestamp.utcnow()`

I chose to leave these while addressing the actual datetime deprecation issue.

PR #101, opened 2025-12-11
Drop support for Python 3.9, add 3.13 to matrix tests

maxinelasp:
Completes #52. 

The tests started failing for 3.9, so this is to drop support now that 3.9 is EOL. The majority of the code should still work for 3.9 but it will no longer be officially supported or tested against.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
     timeout-minutes: 12
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
     defaults:
       run:
         shell: bash
--- a/curryer/compute/elevation.py
+++ b/curryer/compute/elevation.py
@@ -14,7 +14,6 @@ import logging
 import os
 import re
 from pathlib import Path
-from typing import Optional, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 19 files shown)


medley56:
@maxinelasp Since we dropped 3.9, Ruff is now checking pyupgrade against >=3.10 and concluding that we should drop the legacy Union syntax for typehints (and probably other things).

maxinelasp:
Apparently there was a ruff issue that we can resolve now that we are out of 3.9, so unfortunately this PR touches a fair number of files. All those fixes were automatically done by Ruff.

maxinelasp:
As part of this change, I also updated the required checks to include tests for 3.10/11/12/13 (not 9) and pre-commit. These are updates in Github rules.

mmaclay:
Testing this locally with python 3.11 and python 3.13 looks good. All tests pass, but there are new deprecation warnings in 3.13 - 61 warnings for datetime.utcnow() being deprecated:

```
   tests/test_kernels/test_kernels_ephemeris_state.py: 1 warning
   tests/test_integration_ctim.py: 3 warnings
   tests/test_kernels/test_kernels_attitude.py: 3 warnings
   tests/test_kernels/test_kernels_writer.py: 3 warnings
   tests/test_kernels/test_kernels_ephemeris_tle.py: 1 warning
   tests/test_clarreo_eng_scenarios.py: 26 warnings
   tests/test_integration_tsis1.py: 8 warnings
   tests/test_correction/test_monte_carlo.py: 16 warnings
  ```
  
 I think it would make sense to handle that change in this PR and test to make sure it's good with older versions, but obviously not necessary.

Change is in curryer/kernels/writer.py, and should be something like:

```
# Replace:
created=datetime.utcnow().strftime("%Y-%m-%d"),

# With:
created=datetime.now(datetime.UTC).strftime("%Y-%m-%d"),
```
 
 Now I am going to test on cloud9 running some CSDS tests.

maxinelasp:
Now should also address #86

mmaclay (approved):
Looks good and tests all pass for me.

mwatwood-cu (approved):
Woo hoo!

PR #105, opened 2025-12-12
Refactor monte carlo to more generic Correction

### LINKED ISSUE #102: Improve flexibility of GCS (Monte Carlo)
The Correction package in curryer should be flexible and useful for multi-mission applications. While in theory that is the case, it has been developed and tested for specific CPF requirements.

We should explore and add:

1. a simple config example & test case for single parameter optimization (e.g. a timing offset determination in Libera)
2. Improve configuration flexibility:
    * GeolocationConfig Base class, inheriting GeolocationConfig to make a CorrectionClass (not just Monte Carlo, but other methods of sampling and iteration too)

mmaclay:
This refactoring renames Monte Carlo to more generic Correction nomenclature - since the Correction package is more than just Monte Carlo random sampling, it also supports other forms of parameter variation and geolocation optimization.

This change is for monte_carlo.py to correction.py as well as the config: MonteCarloConfig -> CorrectionConfig, and subsequent tests, calls, comments etc.

Addresses #102

### DIFF
--- a/curryer/correction/correction.py
+++ b/curryer/correction/correction.py
@@ -1,7 +1,7 @@
 """
-Mission-Agnostic Monte Carlo Geolocation Correction Pipeline.
+Mission-Agnostic Geolocation Correction Pipeline.
 
-This module provides generic Monte Carlo correction infrastructure for
+This module provides generic correction infrastructure for
 geolocation component sensitivity analysis. It was developed for CLARREO,
 but intended to work with any Earth observation mission through configuration.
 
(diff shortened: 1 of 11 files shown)


curryer/correction/correction.py
```diff
@@ -294,14 +294,14 @@ def _extract_spacecraft_position_midframe(telemetry: pd.DataFrame) -> np.ndarray
...
 
     Args:
         config_path: Path to the JSON configuration file (e.g., gcs_config.json)
 
     Returns:
-        MonteCarloConfig object populated from the JSON file
+        CorrectionConfig object populated from the JSON file
 
     Raises:
         FileNotFoundError: If config file doesn't exist
```
maxinelasp: Need to update logger on ln 316

maxinelasp (approved):
LGTM, assuming you did some kind of search for "monte" to find any lingering comments!

mwatwood-cu (approved):
Approving the improved clarity

PR #121, opened 2026-03-10
Minor bug fix within terrain correction

BStoneLASP:
Minor bug fix within terrain correction that caused a DEM query error under an extreme pointing scenario (intersect ellipsoid with near 90deg off nadir pointing). Required for Clarreo end-to-end testing.

This same logic (excluding high off nadir points from DEM query) occurs later within the terrain correction. The bug was that it wasn't being excluded from the initial query. We never encountered an issue before because usually (every time before) the first query is within the bounding box of the queried DEM data.

### DIFF
--- a/curryer/compute/spatial.py
+++ b/curryer/compute/spatial.py
@@ -908,7 +908,8 @@ def terrain_correct(
     gd_cur_xyz[:, 2] = np.zeros(npts)
 
     # 10) Terrain intersection.
-    gd_cur_h = elev.query(gd_cur_xyz[:, 0], gd_cur_xyz[:, 1])
+    gd_cur_h = np.full(npts, np.nan)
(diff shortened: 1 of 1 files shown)


tmorland77:
Looks good!

tmorland77 (approved):

curryer/compute/spatial.py
mmaclay, line 941: Only question is if you want to add this separate change to this PR: 

`gd_prev_h = gd_cur_h.copy()`

mmaclay (approved):
Looks good to go. There some subtle semantics changes as we now expect gd_cur_h will have NaNs from the beginning, but assuming the testing looks good, this should be an improvement overall.

I added a comment about changing line 941 only if you want to include it in this PR. I think we probably want to reference a copy not the original variable there, but feel free to leave for a later discussion.

mmaclay:
Just updated this branch w/ main to keep this PR from timeout closing automatically.

Issue #60, opened 2025-10-30
Implement surface colatitude


BStoneLASP:
Implement surface point co-latitude within curryer/compute/spatial.

Clarreo geolocation data fields documentation: https://lasp.colorado.edu/galaxy/spaces/LS/pages/288737494/Geolocation+Data+Fields

Related tickets (subsat & subsolar colat): https://github.com/lasp/curryer/issues/58, https://github.com/lasp/curryer/issues/59

PR #157, opened 2026-06-15
LIBSDC-739 Adding new fields and Geometry.py Foundation (PR 1)

### LINKED ISSUE #58: Implement subsatellite lat/lon/colat
Implement subsatellite latitude, co-latitude, and longitude computations within curryer/compute/spatial.

Related ticket (subsolar): https://github.com/lasp/curryer/issues/59

### LINKED ISSUE #59: Implement subsolar lat/lon/colat
Implement subsolar and subsolar latitude, co-latitude, and longitude computations within curryer/compute/spatial.

Related ticket (subsat): https://github.com/lasp/curryer/issues/58

### LINKED ISSUE #65: Implement radius of satellite
Implement radius of satellite from center of earth computation within curryer/compute/spatial.

### LINKED ISSUE #67: Implement earth sun distance
Implement earth sun distance (AU) computation within curryer/compute/spatial.

Either as a "top-of" (e.g., day) or arbitrary time series.

mmaclay:
## Context

The Libera L1B products (and other Level-1 missions) define ~30 geometry/geolocation
ancillary fields — subsatellite/subsolar point, colatitude, satellite radius,
Earth-Sun distance, viewing/solar angles, etc. — that are today filled with `-999`.
With no shared implementation, downstream consumers (libera_rad, libera_cam) have
begun re-deriving these fields independently, and some reimplementations have
diverged (e.g. a subsatellite point computed by intersecting a body axis with the
ellipsoid — attitude-dependent and off by kilometers, versus the correct
position-derived definition).

This PR puts the **generic, mission-agnostic** field math where it belongs — in
curryer — so every mission consumes one correct, tested implementation instead of
hand-rolling its own. It is the **Foundation (PR 1)** of a short series: it lands
the easy single-query, position-derived fields end-to-end and establishes the
registry pattern, frame contract, and fill contract that the later PRs build on.

## Issues addressed

- Closes #58 — Subsatellite point
- Closes #59 — Subsolar point
- Closes #65 — Radius of satellite
- Closes #67 — Earth-Sun distance
- Addresses #60 — Colatitude (the leaf lands here; the surface-footprint
  `surface_colatitude` *field* arrives in the next PR)

## New geometry module and selective-compute framework

* Added the new `geometry.py` module, which implements:
  - Pure, vectorized math-only leaf functions for geometric computations such as
    `sc_radius`, `colatitude`, `subobserver_point`, and `earth_sun_distance`.
    These are importable and composable on their own — the building blocks a
    mission uses for custom fields.
  - A registry-based system mapping each output field to its minimal set of
    required SPICE-derived providers, ensuring each provider is queried only once
    per request (no redundant SPICE passes when several fields share an input).
  - The `GeometryData` class, which offers APIs to retrieve geometric data as
    either a `pandas.DataFrame` or as typed numpy arrays, with robust handling of
    SPICE coverage gaps (NaN propagation, documented as a per-field fill contract
    so downstream maps cleanly onto the product `_FillValue`).
  - Mission genericity: the observing body is the only mission input; no
    spacecraft/instrument names are hardcoded. All ECEF fields share one
    configured Earth-fixed frame (ITRF93 by default).
* Registered the new `geometry` module in the `curryer.compute` package's
  `__init__.py`, updating `__all__` to include it.

## Testing and validation

* Added a comprehensive test suite in `tests/test_compute/test_compute_geometry.py`
  that covers:
  - Unit tests for pure math leaf functions, ensuring correctness without SPICE
    dependencies.
  - Orchestration tests for the `GeometryData` class using fake providers to
    verify minimal querying, field selection, and per-field NaN
  - End-to-end integration tests using real SPICE kernels to validate the full
    provider paths and physical sanity of computed values.

## What's next (follow-on PRs)

This Foundation deliberately scopes to the fields with no open
on top:

- **PR 2** — composed-but-ready fields: instrument boresight and surface-footprint
  colatitude (adds the attitude/boresight/surface providers), p
  `frame_euler` primitive that lets downstream stop open-coding scalar
  `pxform`→`m2eul` loops.
- **Later** — convention-gated fields (along/cross-track, relative azimuth,
  cone/clock, gimbal az/el) and the J2000/ECEF frame-specific s
  fields, once their conventions and kernels are confirmed.

### DIFF
--- a/curryer/compute/__init__.py
+++ b/curryer/compute/__init__.py
@@ -1,6 +1,6 @@
 """Methods for computing spatial and pointing datasets.
 
-@author: Brandon Stone
+@author: Brandon Stone, Matthew Maclay
 """
 
 from . import constants, elevation, ephemeris, geometry, pointing, spatial
--- a/curryer/compute/__init__.py
+++ b/curryer/compute/__init__.py
@@ -3,6 +3,6 @@
 @author: Brandon Stone
 """
 
-from . import constants, elevation, ephemeris, pointing, spatial
+from . import constants, elevation, ephemeris, geometry, pointing, spatial
 
-__all__ = ["constants", "elevation", "ephemeris", "pointing", "spatial"]
+__all__ = ["constants", "elevation", "ephemeris", "geometry", "pointing", "spatial"]
--- a/curryer/compute/geometry.py
+++ b/curryer/compute/geometry.py
@@ -289,6 +289,14 @@ _FIELDS = {
 }
 
 
+# Providers needing only ephemeris (position) coverage -- no attitude (CK) or
+# instrument FOV. The default field set is every field computable from these alone,
+# so a no-`fields` request never implicitly pulls in attitude-derived fields (added
+# by later field groups) that would require additional kernels.
+_EPHEMERIS_PROVIDERS = frozenset({"sc_position", "sun_position"})
+_DEFAULT_FIELDS = tuple(name for name, field in _FIELDS.items() if field.providers <= _EPHEMERIS_PROVIDERS)
+
+
 class GeometryData(abstract.AbstractMissionData):
     """Selective geometric data-field server.
 
@@ -321,7 +329,7 @@ class GeometryData(abstract.AbstractMissionData):
 
     """
 
-    DEFAULT_CADENCE = constants.ATTITUDE_TIMESTEP_USEC
+    DEFAULT_CADENCE = constants.EPHEMERIS_TIMESTEP_USEC
 
     def __init__(self, observer, microsecond_cadence=None, earth="EARTH", sun="SUN", earth_frame=None):
         microsecond_cadence = self.DEFAULT_CADENCE if microsecond_cadence is None else microsecond_cadence
@@ -339,19 +347,35 @@ class GeometryData(abstract.AbstractMissionData):
         return tuple(_FIELDS)
 
     def _resolve_fields(self, fields):
-        """Validate the requested fields, defaulting to all registered."""
+        """Validate the requested fields, defaulting to the ephemeris-only set."""
         if fields is None:
-            return list(_FIELDS)
+            return list(_DEFAULT_FIELDS)
         unknown = [name for name in fields if name not in _FIELDS]
         if unknown:
             raise KeyError(f"Unknown geometry field(s): {unknown}. Available: {self.available_fields()}")
         return list(fields)
 
     def _gather_providers(self, fields, ugps_times):
-        """Query the minimal set of providers for ``fields``, once each."""
-        needed = set().union(*(_FIELDS[name].providers for name in fields))
-        logger.debug("Querying providers %s for fields %s", sorted(needed), fields)
-        return {key: _PROVIDERS[key](ugps_times, self) for key in needed}
+        """Query the minimal set of providers for ``fields``, once each.
+
+        Providers are evaluated in a stable (sorted) order so runs are
+        reproducible. A provider that returns all-NaN over the whole grid is the
+        documented missing-kernel signal, so it is logged as a warning.
+        """
+        needed = sorted(set().union(*(_FIELDS[name].providers for name in fields)))
+        logger.debug("Querying providers %s for fields %s", needed, fields)
+        providers = {}
+        for key in needed:
+            values = _PROVIDERS[key](ugps_times, self)
+            if np.size(values) and np.all(np.isnan(np.asarray(values, dtype=float))):
+                logger.warning(
+                    "Provider %r returned all-NaN over %d time(s); the required kernel is likely "
+                    "unfurnished for this interval.",
+                    key,
+                    len(ugps_times),
+                )
+            providers[key] = values
+        return providers
 
     @abstract.log_return()
     def get_geometry(self, ugps_times: np.ndarray, fields: list[str] | None = None) -> pd.DataFrame:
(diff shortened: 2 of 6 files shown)


curryer/compute/__init__.py
```diff
@@ -3,6 +3,6 @@
 @author: Brandon Stone
```
mwatwood-cu: Seems like you should add your name here @mmaclay

curryer/compute/geometry.py
```diff
@@ -0,0 +1,362 @@
+"""Selective geometric data-field computation.
+
+This module computes the geolocation/geometry ancillary data fields the Level-1
```
mwatwood-cu: Probably don't need the Level-1 verbiage in curryer, unless you are explicitly calling out Level-1 products for specific NASA missions

curryer/compute/geometry.py
```diff
@@ -0,0 +1,362 @@
...
+        ``"EARTH"`` / ``"SUN"``; overridable so no body name is fixed in code.
+    earth_frame : str or spicierpy.obj.Frame, optional
+        Earth-fixed (ECEF) reference frame. Default ``spatial.EARTH_FRAME``
+        (ITRF93).
+
+    """
+
+    DEFAULT_CADENCE = constants.ATTITUDE_TIMESTEP_USEC
+
+    def __init__(self, observer, microsecond_cadence=None, earth="EARTH", sun="SUN", earth_frame=None):
```
medley56, line 334: What happens if a user wants fields calculated on non-uniform time-steps? I don't think it's a good assumption that everything will be smooth curves, so after the fact interpolation is not guaranteed to be a safe option and you'll lose precision that would need to be incorporated into error estimates later on.

I guess I'd expect to be allowed to pass a list of specific times (in some format).

tests/test_compute/test_compute_geometry.py
```diff
@@ -0,0 +1,254 @@
...
+import pytest
+
+from curryer import meta, spicetime, spicierpy, utils
+from curryer.compute import constants, geometry
+
+logger = logging.getLogger(__name__)
+utils.enable_logging(extra_loggers=[__name__])
+
+
+class TestGeometryLeaves:
```
medley56, line 25: I like that you are testing these with known, obvious values!

tests/test_compute/test_compute_geometry.py
```diff
@@ -0,0 +1,254 @@
...
+        assert counter == {"sc_position": 1}
+
+    def test_get_geometry_multiple_fields_union_of_providers(self):
+        counter = {}
+        sc_pos = np.array([[7000.0, 0.0, 0.0], [0.0, 7000.0, 0.0]])
+        sun_pos = np.array([[1.5e8, 0.0, 0.0], [1.5e8, 1.0e6, 0.0]])
+        geo = self._build()
+        fakes = _fake_providers(counter, sc_position=sc_pos, sun_position=sun_pos)
+        with patch.dict(geometry._PROVIDERS, fakes):
+            df = geo.get_geometry(self.UGPS, fields=["sc_position", "subsolar"])
```
medley56, line 120: I'm noticing that there is a lot of magic naming of the fields and the user needs to know the Curryer name of the field they want to calculate. Without going directly into the code, I'm not sure how anyone could know that. 

I think we might need a user documentation update/addition.

medley56 (approved):
Approving but I would like to see an accompanying update to the user docs because the field naming and usage of this module requires some special knowledge of its internals. Also, I'm not sure that calculating all the values on a linspace grid is always good enough. I think we need to support a set of query times, which would then be the independent variable for all fields being calculated.

curryer/compute/geometry.py
mmaclay: Thanks for this. I just clarified this in the docstrings: the times you pass are the independent variable. get_geometry/get_vectors evaluate every field at exactly the ugps_times array (arbitrary, non-uniform is fine), with no resampling or interpolation. microsecond_cadence/get_times is an optional uniform-grid convenience and doesn't constrain which times are queried. So passing a list of specific times already works as you'd expect, no precision is lost to after-the-fact interpolation.

tests/test_compute/test_compute_geometry.py
mmaclay: Yeah this is a good point. Latest push addresses this in two ways: (1) the cryptic column names are now full words and (2) the module docstring now carries an "Available fields" reference listing every field and its columns, alongside the existing GeometryData.available_fields(). 

Do you think it's worth adding and planning a programmatic describe() accessor (columns + per-field description) to make the catalog more self-documenting?

curryer/compute/__init__.py
mmaclay: Added!

curryer/compute/geometry.py
mmaclay: Good call. Dropped the level-1 framing for more generic language as we want.

tests/test_compute/test_compute_geometry.py
mwatwood-cu: Another option would be similar to libera_utils setting up a constants.py or similar that took these hard coded strings and made them enums so they could be imported and used

mwatwood-cu (approved):
I like how this looks. I don't need to re-review, I do think moving the strings that determine which methods get called to an enum class is a nice way to go in the long run, but I'm happy to have the updated documentation here be a first pass

tests/test_compute/test_compute_geometry.py
mmaclay: I'm going to add the ENUM Class in PR 159 (PR 2 of this series)

PR #159, opened 2026-06-17
LIBSDC-739 Add Geometry boresight & surface colatitude (PR 2)

### LINKED ISSUE #60: Implement surface colatitude

mmaclay:
This PR stacks on the [Geometry Foundation PR1 ](https://github.com/lasp/curryer/pull/157) which should be reviewed first. This PR adds the first attitude-derived geometry fields.

### Adds
- **`boresight`** — instrument boresight as a unit vector in ECEF (ITRF93), via the
  tested `spatial.instrument_pointing_state` (resolves the instrument frame from the
  observing body, maps SPICE data gaps to NaN).
- **`surface_colatitude`** — colatitude of the boresight's ellipsoid footprint.

### Closes
- Closes #60 — Colatitude (surface-footprint field; the Foundation landed the
  `colatitude` leaf).

### Efficient by construction
`surface_colatitude` depends on `{boresight, sc_position}` and computes the footprint
with the math-only `ray_intersect_ellipsoid` leaf — no SPICE of its own. `sc_position`
(Foundation) is the shared ray origin, so requesting both fields costs one attitude
query + one shared ephemeris query, never a duplicate pass.

### Tests
27 geometry tests + the compute suite pass, including a real-kernel integration test
(CPRS_HYSICS): boresight comes out finite and unit, footprint colatitude in [0, 180].
ruff clean.


curryer/compute/geometry.py
```diff
@@ -210,9 +210,47 @@ def _provider_sun_position(ugps_times, ctx):
...
+
+    @spicierpy.ext.spice_error_to_val(err_value=None, disable=not ctx.allow_nans)
+    def _resolve_pointing():
+        from_frame = spicierpy.obj.Body(ctx.observer, frame=True).frame.name
+        to_frame = spicierpy.obj.Frame(ctx.earth_frame).name
+        boresight = spicierpy.ext.instrument_boresight(ctx.observer, norm=True)
+        return from_frame, to_frame, boresight
+
+    pointing, _ = _resolve_pointing()
+    if pointing is None:  # missing IK/FOV or frame -> NaN-fill (allow_nans only).
```
medley56: Is this really the behavior you want? I feel like this could surface as a confusing behavior of someone misconfigures their kernel pool and just silently ends up with NaNs everywhere. Should we at least issue a warning? Or maybe the decorator actually does that already?

curryer/compute/geometry.py
```diff
@@ -210,9 +210,47 @@ def _provider_sun_position(ugps_times, ctx):
...
 
 
+# The boresight is a pure attitude quantity: the IK boresight rotated from the
+# instrument frame into ``earth_frame`` by ``pxform`` per sample (``pxform`` has no
+# vectorized override, so it loops). It queries no ephemeris -- position comes from
+# the ``sc_position`` provider -- so requesting the boresight beside the position
+# fields costs one attitude pass plus one shared ephemeris pass, never a duplicate.
+# The SPICE-error guard covers the lookup and the rotation (but no ephemeris), so a
+# NaN row tracks an attitude gap alone, independent of position coverage.
+def _provider_boresight(ugps_times, ctx):
```
medley56, line 265: This needs a docstring that describes the parameters.

curryer/compute/geometry.py
```diff
@@ -242,9 +280,34 @@ def _provider_sun_position(ugps_times, ctx):
...
 
 
+# Providers available for any observing body (queried from ephemeris alone).
+_EPHEMERIS_PROVIDERS = frozenset({"sc_position", "sun_position"})
+
+# Default field set for ``fields=None``: the fields needing only ephemeris, so
+# ``get_geometry()`` is valid for any observer and skips the per-sample attitude
+# loop. Attitude/instrument fields (needing the boresight provider, hence an
+# instrument FOV) are opt-in. Derived from the providers, so it self-maintains.
+_DEFAULT_FIELDS = tuple(name for name, field in _FIELDS.items() if field.providers <= _EPHEMERIS_PROVIDERS)
```
medley56, line 355: What does the `<=` comparison mean for a frozenset? I don't understand this behavior. Is this basically defaulting to only fields that are possible to calculate from ephemeris only? Why is that the default? If a user forgets to furnish an SPK, the ephemeris calculations will fail. If they forget to furnish a CK, the pointing calcuations will fail. Are they so different?

medley56 (approved):
Mostly questions with one request for better docstrings.

curryer/compute/geometry.py
mmaclay: good catch! I added a check in `_gather_providers` so now when a provider returns all-NaN over the whole request it logs a warning naming the provider and pointing at kernel coverage/pool config.  Fill behavior unchanged & allow_nans=False still raises. Also added  test_all_nan_provider_logs_warning / test_partial_nan_provider_does_not_warn.

curryer/compute/geometry.py
mmaclay: Updated this docstring!

curryer/compute/geometry.py
mmaclay: I expanded the inline comment to clarify this a bit:

 <= is the frozenset subset test: a field defaults in when its providers don't include the ephemeris providers. 
 i.e. it's computable from position (an SPK) alone, which every observer has. You're right that this isn't a "definability" line and that missing-SPK vs missing-CK are symmetric coverage failures and a spacecraft does have its own attitude. The reason the attitude fields are opt-in is just extra requirements: boresight (and surface_colatitude) needs the spacecraft attitude (CK, to rotate into ECEF) plus an instrument FOV (IK/FK), so defaulting them would force every no-args call to furnish a CK and use an instrument observer. So the default is simply the minimal-requirements subset (position-only); attitude/instrument fields are opt-in.
 
 Let me know if this is not your concern or want to chat more about this.

curryer/compute/geometry.py
```diff
@@ -249,9 +253,62 @@ def _provider_sun_position(ugps_times, ctx):
...
+
+    Returns
+    -------
+    numpy.ndarray
+        Boresight unit vectors in ``ctx.earth_frame``, shape (N, 3). Rows where
+        the attitude/FOV is unavailable are NaN (under ``allow_nans``).
+    """
+    et_times = spicetime.adapt(ugps_times, to="et")
+
+    @spicierpy.ext.spice_error_to_val(err_value=None, disable=not ctx.allow_nans)
```
mwatwood-cu: How do these errors get reported up the chain for visibility to the user? I've had some difficulties in the past seeing errors when they occur from this function and I'd like to ensure users can see error codes when compute calls fail, but are returnable

curryer/compute/geometry.py
```diff
@@ -249,9 +253,62 @@ def _provider_sun_position(ugps_times, ctx):
...
+        to_frame = spicierpy.obj.Frame(ctx.earth_frame).name
+        boresight = spicierpy.ext.instrument_boresight(ctx.observer, norm=True)
+        return from_frame, to_frame, boresight
+
+    pointing, _ = _resolve_pointing()
+    if pointing is None:  # missing IK/FOV or frame -> NaN-fill (allow_nans only).
+        return np.full((et_times.size, 3), np.nan)
+    from_frame, to_frame, boresight = pointing
+
+    @spicierpy.ext.spice_error_to_val(err_value=np.full(3, np.nan), disable=not ctx.allow_nans)
```
mwatwood-cu: Same question here just making sure these errors get propagated upward to the user

mwatwood-cu (changes requested):
I do want to see how the spice errors are propagated upward/outward to the user.

I'm also more convinced looking at the file level docstring that the different options for geometry should be in their own file as enums.

curryer/compute/geometry.py
mmaclay: Updated from your feedback! 

Two mechanisms, switchable via the allow_nans toggle on GeometryData:
 
- Returnable, not silent (default allow_nans=True): a missing IK/FOV, frame lookup, or coverage gap is caught (via the spice_error_to_val decorator) and NaN-filled per the module's fill contract — the call returns instead of raising. But it's surfaced: _provider_boresight logs a logger.warning naming the observer and the SPICE error (e.g. NOFRAMECONNECT), and _gather_providers warns when a provider comes back all-NaN over the whole request, naming the provider and pointing at kernel coverage / pool config (added for medley's related comment).

- Hard errors when you want them (allow_nans=False): the underlying SpiceyError propagates with its SPICE short/code, so a caller who wants exceptions on failure gets them.

- On returnable error codes specifically: the spice_error_to_val(err_flag=…) + SpatialQualityFlags.from_spice_error() machinery (spicierpy.ext / compute.constants) already maps SPICE errors to flag codes for the pointing/intersection calcs. GeometryData currently surfaces via NaN + logging rather than a returnable flag column (kept minimal per discussion), but that machinery is in place if a per-sample quality-flag column on this API would help your workflow — happy to add it.

curryer/compute/geometry.py
mmaclay: covered in above thread. allow_nans=True surfaces recoverable failures as logged warnings + NaN fill (returnable), and allow_nans=False re-raises the SpiceyError with its code.

curryer/compute/constants.py
mwatwood-cu: Might want to update this

mwatwood-cu (approved):
Looks good!

PR #160, opened 2026-06-17
LIBSDC-780 Surface angles  (PR 3)

### LINKED ISSUE #62: Implement relative azimuth surface angle
Implement relative azimuth surface angle computations within curryer/compute/spatial.

Basically instead of having a solar azimuth angle relative to north, and a viewing azimuth angle relative to north (both we already have logic for), this field is a combination of the two (azimuth between solar and viewing angles).

See CERES documentation page 21-22: https://ceres.larc.nasa.gov/documents/collect_guide/pdf/BDS_CG_R3V4.pdf

### LINKED ISSUE #66: Implement cone/clock angle and rate
Implement cone/clock angle and rate computations within curryer/compute/spatial.

mmaclay:
This PR Stacks on [PR-2](https://github.com/lasp/curryer/pull/159), which should be reviewed first. This one adds the surface-geometry angle fields, all pure math over
providers already queried for other fields — goal is **zero new SPICE**.

### Adds
- **`viewing_zenith`** / **`solar_zenith`** — geodetic zenith of the satellite / Sun
  at the boresight ellipsoid intersection, `[0,180]`.
- **`viewing_azimuth`** / **`solar_azimuth`** — satellite / Sun azimuth there,
  clockwise from geodetic North, `[0,360)`.
- **`relative_azimuth`** (#62) — `mod(viewing_az − solar_az + 180, 360)`, the CERES
  BDS R3V4 origin (Sun at 180); lossless `[0,360)`. The `[0,180]` fold is a downstream step.
- **`cone_angle`** — boresight angle off the satellite-to-geocenter (nadir) vector,
  `[0,90]` (CERES SCI-18).

### Efficient by construction
Each angle is `calc_zenith` / `calc_azimuth` over the boresight ellipsoid
intersection (`ray_intersect_ellipsoid`) — the math-only leaves, not
`spatial.surface_angles`. The intersection reuses `boresight` (PR-B); the targets
reuse `sc_position` / `sun_position` (Foundation). So requesting them all costs the
providers already in hand — no redundant ephemeris re-query and no geodetic
round-trip, which is what the current downstream `surface_angles`-twice usage pays.

### Convention (documented once, in the module)
Azimuths clockwise from geodetic North in `[0,360)`; zeniths geodetic, from the
surface normal; degrees. `relative_azimuth` uses the CERES BDS R3V4 origin —
`mod(viewing_az − solar_az + 180, 360)`, so the Sun sits at 180 — and is the
**lossless** unfolded value; the CERES `[0,180]` fold (`min(raa, 360−raa)`) is a
separate, lossy, one-line downstream step. Curryer keeps the unfolded form
deliberately: a wrap/reference shift is reversible, a fold is not, so only the
unfolded value lets every consumer adapt.

### Tests
42 geometry tests pass (incl. a real-kernel integration test cross-checking the
viewing angles against `spatial.surface_angles` — same answer, cheaper) plus the
full compute suite. Covers physical nadir geometry, leaf-composition wiring,
per-field NaN propagation, a non-fold proof for `relative_azimuth`, and a
nadir/off-nadir `cone_angle` check. ruff clean.

### Base
Targets `geometry-boresight-surface`; auto-retargets to `main` as the stack merges.


curryer/compute/geometry.py
```diff
@@ -254,6 +264,45 @@ def _rotate(sample_et):
...
 
 
+# ---------------------------------------------------------------------------
+# Field-layer helpers: pure math over already-queried providers (no SPICE).
+# These compose the spatial leaves the surface-angle fields share.
+# ---------------------------------------------------------------------------
+_FOOTPRINT_KEY = "footprint"  # cache slot in the per-request providers dict
+
+
+def _footprint(providers):
```
medley56: This semantic choice feels very Libera-coded because this isn't really calculating a footprint -- it has no spatial extent. It's really a ground intersection point or maybe even more generic than that if it misses the ellipsoid.

curryer/compute/geometry.py
```diff
@@ -254,6 +264,45 @@ def _rotate(sample_et):
...
+    """ECEF point where the instrument boresight, cast from the S/C position,
+    meets the ellipsoid -- the shared surface point for the surface angles.
+
+    Memoized into the per-request ``providers`` dict: every surface-angle field in
+    a request shares the same footprint, so the ray-cast runs once per request
+    rather than once per field.
+    """
+    footprint = providers.get(_FOOTPRINT_KEY)
+    if footprint is None:
+        footprint = spatial.ray_intersect_ellipsoid(providers["boresight"], providers["sc_position"])
```
medley56: What happens here if the boresight is off-disk? Does it use npedln and return PNEAR and DIST? What does this footprint variable actually contain and does it support the case of off-disk boresight pointing?
https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/spicelib/npedln.html

Keep in mind that future users will not necessarily want the same behavior as Libera. Sometimes missions treat off-points as if they actually hit the planet at PNEAR (that's what we were asked to do for EMUS). Libera will need to filter these values based on the DIST from the ellipsoid to find the "deep space" observations that we use for the per-scan dark calibration.

curryer/compute/geometry.py
```diff
@@ -254,6 +264,45 @@ def _rotate(sample_et):
...
+
+
+def _footprint(providers):
+    """ECEF point where the instrument boresight, cast from the S/C position,
+    meets the ellipsoid -- the shared surface point for the surface angles.
+
+    Memoized into the per-request ``providers`` dict: every surface-angle field in
+    a request shares the same footprint, so the ray-cast runs once per request
+    rather than once per field.
+    """
```
medley56: Docstring doesn't list parameters.

medley56 (approved):
Couple of questions and the docstrings don't list the parameters and return types.

curryer/compute/geometry.py
mmaclay: Good catch! updated to be `_boresight_intersection`

curryer/compute/geometry.py
mmaclay: It's a direct ray intersect ellipsoid and it does _not_ use npedln or return the PNEAR/DIST. I'm keeping this mission agnostic/generic for now (just returning NaN for a miss), but a future feature to add that should be a reasonably easy addition.

curryer/compute/geometry.py
mmaclay: Updated!

curryer/compute/geometry.py
```diff
@@ -312,6 +328,95 @@ def _rotate(sample_et):
...
+        Per-request provider results; uses ``boresight``, ``sc_position``, and
+        ``sun_position`` (via the shared boresight intersection).
+
+    Returns
+    -------
+    numpy.ndarray
+        Relative azimuth in degrees, shape (N,), in [0, 360); NaN where the
+        boresight misses the ellipsoid.
+    """
+    intersection = _boresight_intersection(providers)
```
mwatwood-cu: What about cases like CPRS and the Libera camera where you don't want the boresight, you want some offset pixel vector?

mwatwood-cu: I see now that the providers dictionary does allow for different pointing unit vectors, but this feels like it is close to duplicating other functions in curryer that allow the passing through of pointing vectors from an instrument kernel

tests/test_compute/test_compute_geometry.py
```diff
@@ -425,6 +527,30 @@ def test_get_geometry_all_fields_real_kernels(self):
         self.assertTrue(surface_colatitude.notna().any(), msg="surface colatitude all-NaN over covered times")
         self.assertTrue(surface_colatitude.dropna().between(0, 180).all())
 
+        # Surface angles resolve where the boresight does, in their documented
+        # ranges: zeniths in [0, 180], azimuths in [0, 360).
```
mwatwood-cu, line 578: Do these calculations include none boresight unit vectors?

curryer/compute/geometry.py
```diff
@@ -312,6 +328,95 @@ def _rotate(sample_et):
...
 
 
+# ---------------------------------------------------------------------------
+# Field-layer helpers: pure math over already-queried providers (no SPICE).
+# These compose the spatial leaves the surface-angle fields share.
+# ---------------------------------------------------------------------------
+_INTERSECTION_KEY = "boresight_intersection"  # cache slot in the per-request providers dict
+
+
+def _boresight_intersection(providers):
```
mwatwood-cu: I'm not sold on the name _boresight_intersection if you are allowing and array of "boresights". I'm ok if that is the standard naming, but it feels funky to me.

curryer/compute/geometry.py
```diff
@@ -312,6 +328,95 @@ def _rotate(sample_et):
...
+
+    A boresight that misses the ellipsoid (off-disk / deep-space pointing) yields
+    NaN: ``ray_intersect_ellipsoid`` is a direct ray-ellipsoid intersection, not a
+    nearest-point projection, so it returns no PNEAR/DIST. Memoized into the
+    per-request ``providers`` dict, so the ray-cast runs once per request and is
+    shared by every surface-angle field.
+
+    Parameters
+    ----------
+    providers : dict
```
mwatwood-cu: I don't love passing in a dictionary that requires specific string values as the keys. Especially since this is only two parameters, I think they should be passed in separately and be more explicit in the doc string so is is crystal clear what this method does (even though it is a private/internal method)

curryer/compute/geometry.py
```diff
@@ -312,6 +328,95 @@ def _rotate(sample_et):
...
+        misses the ellipsoid.
+    """
+    intersection = providers.get(_INTERSECTION_KEY)
+    if intersection is None:
+        intersection = spatial.ray_intersect_ellipsoid(providers["boresight"], providers["sc_position"])
+        providers[_INTERSECTION_KEY] = intersection
+    return intersection
+
+
+def _relative_azimuth(providers):
```
mwatwood-cu: Should this be called name more explicitly to be a CERES convention or maybe a flag that identifies it as such? I don't know of other use cases for relative azimuth angle, so maybe this isn't necessary, but I wanted to point it out.

mwatwood-cu (changes requested):
I'm not convinced this will work smoothly for applications like the Libera camera that will require passing through many pixel pointing vectors. Maybe just some more explicit tests and function naming would clear this up for me.

I also would like to avoid the dict passing even for internal methods, but I'm ok being overruled if that is your preferred style

curryer/compute/geometry.py
mmaclay: Good questions here, this helped me narrow the scope a bit (at least for now).

so now `GeometryData` is for spacecraft-level quantities plus angles referenced to the instrument's single IK boresight, one row per time. All per-pixel / offset-pixel geometry (WFOV pixel vectors, a CPRS offset pixel) is a (time, pixel) problem, served by the existing `spatial.compute_ellipsoid_intersection(custom_pointing_vectors=...)` which is vectorized, quality-flagged, and already what libera_cam uses for the focal plane.

So this composes the same math leaves (ray_intersect_ellipsoid/calc_zenith/calc_azimuth) for the one boresight rather than duplicating that path. I updated the GeometryData docstring, and the per-request container is now a typed `_ProviderResults` whose boresight attribute is explicitly the IK boresight. A single non-boresight reference vector here should be a small additive extension (one vector, not the array) if a mission ever needs it.

tests/test_compute/test_compute_geometry.py
mmaclay: no this is just for fields derived from the instruments IK boresight

curryer/compute/geometry.py
mmaclay: Refactored to _ProviderResults.boresight_intersection (a cached_property), which is the single boresight's intersection. For the "array of boresights": the (N, 3) shape is the time axis, the instrument's one boresight expressed in ECEF at each of N times (same N as every field), not N distinct foresights. So multiple pointing vectors (pixels/offsets) should continue to go in the separate compute_ellipsoid_intersection path

curryer/compute/geometry.py
mmaclay: providers dict is now a typed _ProviderResults dataclass with named attributes (p.sc_position, p.boresight, …), so fields read attributes instead of string keys. hopefully improves observability here

curryer/compute/geometry.py
mmaclay: Docstring includes a ceres reference, so going to leave this as the generic relative_azimuth for now.

mmaclay:
Updated to clarify the role of `GeometryData` compared to the existing compute_ellipsoid_intersection, which should still be used for multi-pixel cases. The geometry module has the leaves to do efficient spacecraft and instrument observation level fields and calculations, but not offset pixels currently. This could be added with a more generic pointing_vector in the future.

Issue #64, opened 2025-10-30
Implement satellite component attitude querying


BStoneLASP:
Implement satellite component (e.g., gimbal azimuth and elevation) attitude querying / conversion (rad/deg) within curryer/compute/spatial.

Simply a wrapper around the SPICE attitude query with standard input arg formats (e.g., ugps_times) and conversion options (e.g., deg/rad).

Clarreo geolocation data fields documentation: https://lasp.colorado.edu/galaxy/spaces/LS/pages/288737494/Geolocation+Data+Fields

PR #161, opened 2026-06-18
Add standalone frame_euler function for Euler angle conversion between frames

### LINKED ISSUE #158: Add a frame_euler helper so we stop hand-rolling pxform→m2eul
The Euler-angles-between-two-frames pattern (pxform → m2eul → degrees) only exists in curryer today buried inside spatial.instrument_pointing_state (spatial.py:277). That function needs an instrument Body with ephemeris and also computes pointing vectors, so it's unusable when you just want the angles between two coordinate frames — which is exactly what gimbal/motor azimuth and elevation are. As a result the pattern keeps getting copy-pasted downstream.

We should add a small generic helper in compute/spatial.py:

frame_euler(from_frame, to_frame, ugps_times, sequence=(1,2,3), degrees=True, allow_nans=True)

It loops pxform/m2eul over the time grid (there's no vectorized pxform — only query_ephemeris is vectorized — so per-time evaluation is unavoidable and matches what curryer already does), wraps errors with the existing `spicierpy.ext.spice_error_to_val` so gaps become NaN + a quality flag instead of silent failures, and returns the angles indexed by uGPS. Nothing mission-specific goes in — frame names stay with the caller.

One caveat worth a docstring line: at high output rates (e.g. camera at 100 Hz) callers should evaluate this on a coarse cadence and interpolate (angles are smooth; unwrap azimuth first), using the existing AbstractMissionData.get_times(cadence) pattern rather than calling SPICE per output sample.

Where this needs to land / get rewired:

- curryer — add the helper, and refactor instrument_pointing_state to call it (no behavior change, just one code path).
- libera_rad — calculate_azimuth_elevation_for_timestamps in PR #20 (libera_rad/geolocation.py) currently open-codes two pxform+m2eul calls per sample inside a Python loop with a bare except Exception: continue. That's the production consumer; it rewires to frame_euler and drops the loop and the swallow-everything except. Coordinate timing so #20 merges as explicitly interim.
- libera_utils — tests/integration/test_tier0_kernel.py:270–278 is actually where the LIBERA_BASE→AZ→EL convention lives (az = m2eul[2] of BASE→AZ, el = m2eul[0] of AZ→EL). PR #20's code says it "matches" this test, so that test is our only source of truth for the axis mapping. We should move that convention into frame_euler's tests and have the tier-0 test assert against the helper, so there's one definition instead of a test and a production copy that can drift.
- libera_cam — not affected; it doesn't compute motor angles (no pxform/m2eul in the repo).

Net: one tested primitive in curryer, one downstream rewire (libera_rad), and the convention pinned in code rather than in a test.

mmaclay:
This pull request introduces a new utility function for computing Euler angles between reference frames over time, along with comprehensive unit tests to validate its behavior and integration. The main focus is on enabling flexible, robust extraction of frame-to-frame Euler angles using SPICE kernels, and ensuring correctness through both unit and integration-style tests.

### potential usage and downstream consumers
the consumers below are mostly future or illustrative, not yet wired:

1. In-repo (curryer spatial.py) — instrument_pointing_state already does this same math inline (m2eul(pxform(...), 1, 2, 3) → degrees). frame_euler is the standalone extraction of it. Folding the inline copy onto frame_euler is an optional later consolidation (rides the deferred spatial.py breakup) — not required for frame_euler to be complete.
2. Downstream (libera_rad) — the stage-by-stage motor Az/El loop on feature/geo-subsatellite-azel (geolocation.py:225–230) is the eventual rewire target. Caveat: that branch is an unmerged example, not production — it illustrates the manual pxform→m2eul pattern frame_euler will replace once curryer releases the primitive and libera_rad bumps the pin. LIBERA frames stay downstream; the conversion comes from curryer.
3. Future curryer field — gimbal Azimuth/Elevation (#64) — not yet implemented. It's in the convention-gated group, still pending the frame pair / Euler sequence and corrected-vs-raw decisions. When built, it will be a thin wrapper over frame_euler, and the libera_rad branch above is the candidate definition to reconcile against.

So this adds frame_euler as a ready building block now. Az/El (both the curryer field and the downstream rewire) is still to come — gated on conventions and on the downstream branches actually merging — so today's PR adds the reusable primitive ahead of those consumers, not the Az/El fields themselves.

**New Euler angle computation utility:**

* Added the `frame_euler` function to `curryer/compute/spatial.py`, which computes the Euler angles required to rotate one reference frame into another at specified times. This function is generic over frame pairs, supports custom axis sequences, handles both degrees and radians, and can optionally fill missing data with NaNs when SPICE coverage is incomplete.

**Comprehensive testing for new functionality:**

* Added a suite of unit tests in `tests/test_compute/test_compute_spatial.py` to verify:
  - Input validation for axis sequences (must be three axes, no repeats, valid range).
  - Correct output in both degrees and radians, with appropriate column names and index.
  - NaN-filling behavior for attitude gaps when `allow_nans=True`, and error propagation otherwise.
  - Consistency with the existing `instrument_pointing_state` function over real SPICE kernels.
  - Ability to reconstruct the original rotation matrix from the returned Euler angles, verifying mathematical correctness.

### DIFF
--- a/curryer/compute/spatial.py
+++ b/curryer/compute/spatial.py
@@ -323,7 +323,85 @@ def instrument_pointing_state(
     return pnt_data, sc_data, qf_data
 
 
-def frame_euler(
+def frame_to_frame_rotation(
+    from_frame: int | str | spicierpy.obj.Frame,
+    to_frame: int | str | spicierpy.obj.Frame,
+    ugps_times: np.ndarray,
+    allow_nans=False,
+) -> np.ndarray:
+    """Rotation matrices taking one reference frame into another over time.
+
+    The per-sample ``pxform(from_frame, to_frame, t)`` rotation stacked over the
+    requested times -- the shared SPICE query behind :func:`frame_to_frame_euler` and
+    :func:`frame_to_frame_quaternion`. Call it directly when the matrix itself is wanted
+    (e.g. to rotate vectors) or to derive several representations from a single query.
+
+    Note
+    ----
+    ``pxform`` has no vectorized SpiceyPy override, so this loops per sample -- an
+    inherent scalar cost, not a missing optimization. For dense time grids, compute on a
+    coarse CK-cadence grid (see ``AbstractMissionData.get_times``) and interpolate rather
+    than calling this at full rate.
+
+    Parameters
+    ----------
+    from_frame, to_frame : str or int or spicierpy.obj.Frame
+        Source and destination reference frames; the returned matrices rotate
+        ``from_frame`` into ``to_frame``. Relevant CK/FK kernels must be loaded.
+    ugps_times : array_like of int
+        One or more times in GPS microseconds.
+    allow_nans : bool, optional
+        Convert recoverable SPICE failures (e.g. attitude gaps) into all-NaN matrices.
+        Default False raises instead.
+
+    Returns
+    -------
+    numpy.ndarray
+        Rotation matrices, shape (N, 3, 3). Samples without attitude coverage are
+        all-NaN when ``allow_nans``.
+
+    """
+    from_name = spicierpy.obj.Frame(from_frame).name
+    to_name = spicierpy.obj.Frame(to_frame).name
+    ugps_times = np.atleast_1d(np.asarray(ugps_times))
+    et_times = spicetime.adapt(ugps_times, to="et")
+
+    @spicierpy.ext.spice_error_to_val(err_value=np.full((3, 3), np.nan), disable=not allow_nans)
+    def _rotation(sample_et):
+        return spicierpy.pxform(from_name, to_name, sample_et)
+
+    return np.array([_rotation(sample_et)[0] for sample_et in et_times])
+
+
+def _rotation_rows(matrices, convert, n_columns):
+    """Apply a per-matrix ``convert`` to each rotation, passing NaN matrices through.
+
+    Attitude gaps arrive from :func:`frame_to_frame_rotation` as all-NaN matrices; those
+    rows stay NaN (``convert`` -- a SPICE routine -- is never called on invalid input).
+
+    Parameters
+    ----------
+    matrices : numpy.ndarray
+        Rotation matrices, shape (N, 3, 3).
+    convert : callable
+        Maps one (3, 3) matrix to a length-``n_columns`` representation.
+    n_columns : int
+        Width of the converted output.
+
+    Returns
+    -------
+    numpy.ndarray
+        Converted rows, shape (N, n_columns); all-NaN for NaN input matrices.
+
+    """
+    rows = np.full((len(matrices), n_columns), np.nan)
+    for ith, matrix in enumerate(matrices):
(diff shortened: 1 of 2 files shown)


curryer/compute/spatial.py
```diff
@@ -323,6 +323,77 @@ def _query(sample_et):
     return pnt_data, sc_data, qf_data
 
 
+def frame_euler(
```
medley56: Maybe out of scope but it would be cool if this could optionally return Euler angles, quaternions, or rotation matrices.

curryer/compute/spatial.py
```diff
@@ -323,6 +323,77 @@ def _query(sample_et):
...
+        raise ValueError(f"`sequence` may not repeat adjacent axes (m2eul restriction), got: {sequence}")
+
+    from_name = spicierpy.obj.Frame(from_frame).name
+    to_name = spicierpy.obj.Frame(to_frame).name
+    ugps_times = np.atleast_1d(np.asarray(ugps_times))
+    et_times = spicetime.adapt(ugps_times, to="et")
+
+    @spicierpy.ext.spice_error_to_val(err_value=np.full(3, np.nan), disable=not allow_nans)
+    def _euler(sample_et):
+        return spicierpy.m2eul(spicierpy.pxform(from_name, to_name, sample_et), *sequence)
```
medley56: To return rotation matrices, you could just omit the m2eul call. I wonder if there is a m2quat call or something similar.

medley56 (approved):
Just one minor suggestion to allow returning multiple possible representations of the rotation.

curryer/compute/spatial.py
mmaclay: I think theres `m2q` - I'll look into this!

curryer/compute/spatial.py
mwatwood-cu: What do you think of `frame_to_frame_euler_angle` as a name? Or something similar, the name feels a bit short to be fully descriptive. Not required, just an idea.

mwatwood-cu (approved):
I like it! I'm with Gavin, could be useful to add more options for return values, but not required. I also request a consideration of a more verbose name for this method

curryer/compute/spatial.py
mmaclay: Good call. Renaming to your suggestion (and using same format for other output formats quaternions etc)

curryer/compute/spatial.py
mmaclay: Adding all three. frame_to_frame_rotation returns the (N, 3, 3) matrices, frame_to_frame_quaternion returns scalar-first q0..q3 via m2q, and frame_to_frame_euler is the renamed Euler form. They all share one pxform query, so any representation (or several) comes from a single SPICE pass.

curryer/compute/spatial.py
mmaclay: added all 3!

mmaclay:
Merging this additive function since there is no previous usage to break. @BStoneLASP Tagging you for visibility here.

PR #165, opened 2026-07-01
Release documentation

maxineofficial:
Add documentation around release and general repo management. Also fixed a few other issues with the docs that I found.


mmaclay (approved):
Looks good! I think some of the copilot review stuff is reasonable to add in, but not necessary.

PR #169, opened 2026-08-12
KM: PR 1 Kernel loading

mmaclay:
This pull request introduces a new utility for inspecting the currently loaded SPICE kernels and adds comprehensive tests to ensure its correctness and flexibility.

**New kernel pool inspection utility:**

* Added a `LoadedKernel` named tuple to `ext.py` to represent metadata about each furnished kernel, including its file path, type, source, and handle.
* Introduced the `loaded_kernels` function in `ext.py`, which queries the SPICE kernel pool directly and returns a list of `LoadedKernel` records, supporting filtering by kernel type.

**Testing and validation:**

* Added `test_loaded_kernels_lists_pool` to verify that `loaded_kernels` accurately reflects the pool state, including correct file listing and type checking.
* Added `test_loaded_kernels_kind_filter` to ensure the `kind` filter works as expected for different kernel types, including type and handle validation.

### DIFF
--- a/curryer/spicierpy/ext.py
+++ b/curryer/spicierpy/ext.py
@@ -100,6 +100,56 @@ class load_kernel:
             raise ValueError("Must specify `kernel` (str) or `all`=True.")
 
 
+class LoadedKernel(typing.NamedTuple):
+    """A kernel currently furnished in the SPICE kernel pool.
+
+    Attributes
+    ----------
+    file : str
+        Kernel file path, as it was furnished.
+    ktype : str
+        SPICE kernel type (e.g. ``"SPK"``, ``"CK"``, ``"PCK"``, ``"TEXT"``, ``"META"``).
+    source : str
+        Name of the source file that furnished this kernel (e.g. a meta-kernel),
+        or an empty string if it was furnished directly.
+    handle : int
+        SPICE integer handle for binary kernels; 0 for text kernels.
+    """
+
+    file: str
+    ktype: str
+    source: str
+    handle: int
+
+
+def loaded_kernels(kind="ALL"):
+    """List the kernels currently furnished in the SPICE kernel pool.
+
+    Queries the kernel pool itself (``ktotal`` / ``kdata``), so the result reflects
+    every furnished kernel regardless of which component loaded it — unlike
+    :attr:`load_kernel.loaded`, which tracks only the kernels loaded through that
+    handle. Useful for verifying furnish state and for pool-wide diagnostics.
+
+    Parameters
+    ----------
+    kind : str, optional
+        SPICE kernel kind filter: one of ``"SPK"``, ``"CK"``, ``"PCK"``, ``"DSK"``,
+        ``"EK"``, ``"TEXT"``, ``"META"``, a space-delimited combination, or ``"ALL"``.
+        Default="ALL".
+
+    Returns
+    -------
+    list of LoadedKernel
+        One record per furnished kernel, in load order.
+
+    """
+    records = []
+    for which in range(spiceypy.ktotal(kind)):
+        file, ktype, source, handle = spiceypy.kdata(which, kind)
+        records.append(LoadedKernel(file=file, ktype=ktype, source=source, handle=handle))
+    return records
+
+
 def object_frame(obj_name, as_id=False):
     """Retrieve frame name/id associated with an object name or id.
 
--- a/curryer/spicierpy/ext.py
+++ b/curryer/spicierpy/ext.py
@@ -3,6 +3,7 @@
 @author: Brandon Stone
 """
 
+import enum
 import logging
 import os
 import shutil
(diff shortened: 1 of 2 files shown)


curryer/spicierpy/ext.py
```diff
@@ -100,6 +100,56 @@ def unload(self, kernel=None, clear=False):
...
+    every furnished kernel regardless of which component loaded it — unlike
+    :attr:`load_kernel.loaded`, which tracks only the kernels loaded through that
+    handle. Useful for verifying furnish state and for pool-wide diagnostics.
+
+    Parameters
+    ----------
+    kind : str, optional
+        SPICE kernel kind filter: one of ``"SPK"``, ``"CK"``, ``"PCK"``, ``"DSK"``,
+        ``"EK"``, ``"TEXT"``, ``"META"``, a space-delimited combination, or ``"ALL"``.
+        Default="ALL".
```
medley56: This is a pretty thin wrapper around `kdata`. I'd like to see this interface be a little friendlier. 

- Create a StrEnum for the valid string values for kernel type ("SPK", "CK" etc).
- Typehint `kind` as `KernelType | list[KernelType]`
- Assemble a list of strings into the space-delimited args as expected by ktotal and kdata.

medley56 (approved):
Minor comment. Take it or leave it but it's an easy change.

curryer/spicierpy/ext.py
```diff
@@ -100,6 +100,56 @@ def unload(self, kernel=None, clear=False):
...
+
+    Returns
+    -------
+    list of LoadedKernel
+        One record per furnished kernel, in load order.
+
+    """
+    records = []
+    for which in range(spiceypy.ktotal(kind)):
+        file, ktype, source, handle = spiceypy.kdata(which, kind)
```
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

`spiceypy.kdata` is wrapped in `@spice_found_exception_thrower`, which strips the trailing `found` flag only while `spiceypy.config.catch_false_founds` is truthy — a documented global that users can flip. With it disabled, `kdata` returns a 5-tuple and this line raises a bare `ValueError: too many values to unpack`.

Curryer never sets that flag today (confirmed by grep), so this is latent rather than a live bug. Cheapest guard:

```python
file, ktype, source, handle = spiceypy.kdata(which, kind)[:4]
```

medley56 (approved):

Clean, well-scoped helper, and the test hardening after the earlier review round is good. Two small comments plus a note on the open `kind` thread. Non-blocking.

**Stack note:** no production caller ships with this — #172 is the first consumer. `leapsecond.are_loaded()` (`curryer/spicetime/leapsecond.py:83`) already hand-rolls this exact loop; converting it would give the PR a real caller and prove the API:

```python
[rec.file for rec in ext.loaded_kernels(kind="text") if sp.getfat(rec.file)[1] == "LSK"]
```

(That file isn't in this PR's diff, hence the note here rather than inline.)

**On the `kind` interface thread:** strongly agree with the `StrEnum` suggestion, and the enum gives a home for a caveat that's currently undocumented — see the reply on that thread.

curryer/spicierpy/ext.py
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

Agreed on the `StrEnum` — and it gives a natural home for a caveat that's currently undocumented anywhere:

SPICE reports *binary* PCKs as `PCK`, but collapses **every** text kernel (FK, IK, LSK, SCLK, **and text PCKs**) into `TEXT`. Verified with both a `.tpc` and a `.bpc` furnished:

```
ktotal('ALL')  -> 2      ALL[0] ktype=TEXT  pck00010.tpc
ktotal('PCK')  -> 1      ALL[1] ktype=PCK   earth_000101_221016_220723.bpc
```

So `kind="PCK"` matches only the binary one, and callers filtering on `rec.ktype == "PCK"` silently miss text PCKs. Distinguishing text-kernel subtypes requires `spiceypy.getfat(rec.file)`, as `leapsecond.are_loaded()` does.

Worth putting on the `KernelType` docstring, since #172 filters on `rec.ktype`.

mmaclay: Good call — `kind` now takes a `KernelType` member, a plain string, or a list of either (`(str, Enum)` rather than `StrEnum`, since we still support 3.10). Your text-vs-binary PCK finding is on the enum docstring with a test pinning it; that one would have bitten someone quietly.

curryer/spicierpy/ext.py
mmaclay: Nice catch on a latent one — sliced to the first four values, with a comment naming the config flag so the slice doesn't look arbitrary.

mmaclay:

The `kind` argument is friendlier now: a `KernelType` member, a plain string, or a list of either. It's `(str, Enum)` rather than `StrEnum` because the floor is still Python 3.10. Your finding that SPICE reports text PCKs as TEXT is on the enum's docstring with a test pinning it, which is the part most likely to save someone downstream. The `kdata` slice guard is in too.

One thing I didn't do: converting `leapsecond.are_loaded()` to use this. It sits in the other chain, and wiring the two together now would couple branches that are deliberately independent — it lands naturally in the kernel manager work, where both exist.

Ready for another look.

PR #170, opened 2026-08-12
KM PR 2: Enhance PCK coverage

mmaclay:
This pull request enhances support for binary PCK (orientation) kernels in the `spicierpy.ext` module, allowing coverage and object queries for these files. It adds handling for binary PCKs keyed on frame class IDs, improves error messages and documentation, and introduces comprehensive tests for the new behavior.

**Support for binary PCK kernels:**

* Updated `kernel_coverage` to support binary PCKs (arch "DAF") by querying coverage using frame class IDs, and to raise clear errors for unsupported text PCKs and invalid body types. [[1]](diffhunk://#diff-2dc9844d34ae893d0e0b16dcaee7dbb8c0aa014106ae52bb78a5f6d0a6a5519cR186-R188) [[2]](diffhunk://#diff-2dc9844d34ae893d0e0b16dcaee7dbb8c0aa014106ae52bb78a5f6d0a6a5519cL203-R208) [[3]](diffhunk://#diff-2dc9844d34ae893d0e0b16dcaee7dbb8c0aa014106ae52bb78a5f6d0a6a5519cL229-R246)
* Updated `kernel_objects` to return frame class IDs for binary PCKs, raise errors for unsupported text PCKs, and clarify usage in the docstring. [[1]](diffhunk://#diff-2dc9844d34ae893d0e0b16dcaee7dbb8c0aa014106ae52bb78a5f6d0a6a5519cL275-R295) [[2]](diffhunk://#diff-2dc9844d34ae893d0e0b16dcaee7dbb8c0aa014106ae52bb78a5f6d0a6a5519cR309-R320)

**Testing improvements:**

* Added new tests for binary PCK coverage and object queries, including segment handling, error cases, and correct output formats.
* Improved test for text PCK error handling to check for updated error messages.

**Documentation and imports:**

* Clarified docstrings to explain handling of frame class IDs and updated test imports to include `spicetime` for time format conversions. [[1]](diffhunk://#diff-2dc9844d34ae893d0e0b16dcaee7dbb8c0aa014106ae52bb78a5f6d0a6a5519cR186-R188) [[2]](diffhunk://#diff-2dc9844d34ae893d0e0b16dcaee7dbb8c0aa014106ae52bb78a5f6d0a6a5519cL275-R295) [[3]](diffhunk://#diff-c726870175bcd81147d4a7dca07779cca14d6e4273508eb5c6c6a39f7fa1d722L14-R14)


curryer/spicierpy/ext.py
medley56: This kind of typing might work in a docstring but probably better to write `str | Path`. Also, missing typehints in the function signature itself.

tests/test_spicierpy/test_spicierpy_ext.py
```diff
@@ -262,6 +262,70 @@ def test_kernel_objects(self):
             ext.kernel_objects(tmp_file)
         self.assertIn("Unknown or unexpected kernel type: 'SCLK' from", raised.exception.args[0])
 
+    def _write_test_pck(self, filename, classid=3000, segments=((0.0, 86400.0),)):
```
medley56, line 283: I love this way of testing things! Writing test data dynamically is awesome.

medley56 (approved):
Just noticed some typehinting missing.

curryer/spicierpy/ext.py
```diff
@@ -226,10 +231,26 @@ def kernel_coverage(filename, body, as_segments=False, to_fmt="ugps"):
...
+        # Text PCKs (arch "KPL") carry constants, not time coverage.
+        if arch != "DAF":
+            raise NotImplementedError(f"Text PCK kernels carry constants, not time coverage: {filename!r}")
+        if isinstance(body, Body | Frame):
+            # Coverage is keyed on the frame *class* ID (e.g., 3000 for ITRF93),
+            # not the frame ID (13000): resolve through the frame info, going
+            # via the body's associated frame for Body inputs. Requires the
+            # frame definitions to be loaded.
+            frame_id = spiceypy.cidfrm(body.id)[0] if isinstance(body, Body) else body.id
+            idcode = spiceypy.frinfo(frame_id)[2]
```
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

**Blocking.** `frinfo(...)[2]` is the PCK class ID only when the frame is frame-class 2; the class field `frinfo(...)[1]` is ignored here.

Verified against kernels already in `data/generic`:

- `Frame("ITRF93")` -> class 2, class ID 3000 — works.
- `Frame("MOON_PA")` and `Body("MOON")` -> **31000**, a class-4 TK alias.
- But `moon_pa_de421_1900-2050.bpc` contains **31006**.

So the call dies with a misleading `No data for body [Frame(MOON_PA)] was found in the kernel containing IDs=[31006]`. Those kernels are referenced from every meta-kernel in `tests/data/*/`, so this is the mainline lunar case, not an edge case.

Suggest resolving the class properly — while `frclss == 4`, follow `spiceypy.tkfram(frame_id)[1]`; then require `frclss == 2` and use `clssid`, raising a clear error otherwise. Verified: `tkfram(31000)[1]` -> `31006`, and `frinfo(31006)` -> `(301, 2, 31006)`, which matches the kernel.

curryer/spicierpy/ext.py
```diff
@@ -226,10 +231,26 @@ def kernel_coverage(filename, body, as_segments=False, to_fmt="ugps"):
...
+        elif isinstance(body, int):
+            idcode = body
+        else:
+            raise TypeError(
```
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

The "no name lookup" rationale is true of the class *ID*, but not of the name the caller passes: `Frame("ITRF93")` is accepted and resolves fine with no FK loaded, while the bare string `"ITRF93"` raises.

The SPK and CK branches both coerce strings via `Body(...)`/`Frame(...)`, the function's own docstring still promises `body : int or str`, and existing callers pass names (`tests/test_kernels/test_kernels_attitude.py:82`, `..._ephemeris_state.py:61`). This branch is the odd one out.

Suggest folding strings into the same path (`body = Frame(body)`, falling back to `Body(body)` on a name-lookup failure) and keeping `TypeError` only for genuinely unusable input.

curryer/spicierpy/ext.py
```diff
@@ -291,6 +313,18 @@ def kernel_objects(filename, as_id=False):
...
+    # Orientation (binary PCK) kernel: contains frame *class* IDs (e.g., 3000
+    # for ITRF93), which have no generic name lookup in SPICE.
+    elif ktype == "PCK":
+        # Text PCKs (arch "KPL") carry constants, not object/frame segments.
+        if arch != "DAF":
+            raise NotImplementedError(f"Text PCK kernels carry constants, not object segments: {filename!r}")
+        ids_cell = spiceypy.support_types.SPICEINT_CELL(1000)
+        spiceypy.pckfrm(filename, ids_cell)
+        objs = tuple(int(ids_cell[i]) for i in range(spiceypy.card(ids_cell)))
+        if not as_id:
```
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

The `pckfrm` query runs and `objs` is built, then the default `as_id=False` throws it all away with `NotImplementedError` — so the default call signature is unusable for binary PCKs, and polymorphic callers (including #172's coverage module) must special-case PCK before they can call this at all.

Suggest either returning the integer class IDs regardless of `as_id` (documented, since there is no name form), or raising *before* the `pckfrm` call so the intent is obvious and the work isn't wasted.

tests/test_spicierpy/test_spicierpy_ext.py
medley56, line 283: 🤖 *Comment generated by Claude and reviewed by Gavin*

`_write_test_pck` builds a class-3000 kernel, which is exactly the one case where the naive `frinfo[2]` mapping happens to be correct — so the tests can't catch the alias-frame issue flagged on `ext.py`.

The repo already ships real binary PCKs that would have: `data/generic/earth_000101_221016_220723.bpc` and `moon_pa_de421_1900-2050.bpc` (the Moon one needs `moon_assoc_pa.tf` + `moon_080317.tf` loaded). Suggest adding a case over those.

medley56 (changes requested):

Right approach, and documenting that binary-PCK coverage is keyed on the frame *class* ID is a real improvement. The dynamic test-kernel construction is nice.

Requesting changes on one item: the class-ID resolution mis-resolves the Moon PA kernels this repo ships and every test meta-kernel references.

Worth noting the earlier resolved thread diagnosed this correctly (`body.id` is not the class ID), but the fix only covers frame-class 2 — the alias case is still open.

curryer/spicierpy/ext.py
mmaclay: Done — real signatures on all three now, with `str | Path` where you'd expect.

tests/test_spicierpy/test_spicierpy_ext.py
mmaclay: Thanks! Though the note you left below is the fair counterpoint: built fixtures only test the shape I thought to build, which is exactly how the moon alias bug got through. Real kernels are in there alongside them now.

curryer/spicierpy/ext.py
mmaclay: Best catch of the review — confirmed exactly as you described, and the `Body` path was broken the same way. There's a shared `frame_class_id` helper now that follows the alias chain and refuses anything that isn't a PCK frame, tested against the real moon and earth kernels.

curryer/spicierpy/ext.py
mmaclay: You're right that this branch was the odd one out. Names go through the same path now, so the frame, the alias, and the body name all work; anything unresolvable gets an error telling you to load the frame kernel.

curryer/spicierpy/ext.py
mmaclay: Agreed, that ordering was silly — it raises before doing the work now.

tests/test_spicierpy/test_spicierpy_ext.py
mmaclay: Exactly right, and that's why the bug survived to review. Added cases over both kernels you named, plus the error paths.

mmaclay:

The moon fix is the substantive one. Class-ID resolution now follows the alias chain properly and refuses anything that isn't a PCK frame, so `MOON_PA`, `MOON_PA_DE421`, `MOON`, and the `Body` form all resolve to what the kernel actually contains. Names work in the PCK branch now like they already did for SPK and CK, and `kernel_objects` raises before doing work it throws away.

You were right that the built fixtures couldn't catch this — there are tests over the real moon and earth kernels now, plus the error paths.

The resolution lives in a shared `frame_class_id` helper, which also let the coverage module in #172 drop its duplicate copy.

Ready for another look.

medley56 (approved):

PR #172, opened 2026-08-12
KM PR 3 Add coverage module for SPICE kernel coverage aggregation

mmaclay:
This pull request introduces a new `coverage` module to the `curryer.kernels` package, providing comprehensive utilities for SPICE kernel coverage analysis. It also adds a suite of unit tests to ensure the correctness of these new features and updates the package's exports to include the new module.

**New kernel coverage functionality:**

* Added `curryer/kernels/coverage.py`, which provides functions to:
  - Compute the union of coverage windows for a target object across kernels (`object_coverage`).
  - Find windows where all specified targets are simultaneously covered (`valid_window`).
  - Identify uncovered sub-ranges within a requested window (`coverage_gaps`).
  - Roll up per-kernel, per-object coverage windows for diagnostics (`coverage_rollup`).
  - Compute overlapping coverage windows for a target between two kernels (`pairwise_overlap`).

**Testing and integration:**

* Added `tests/test_kernels/test_kernels_coverage.py`, a comprehensive unit test suite that exercises all major functionalities of the new `coverage` module, including edge cases and error handling.
* Updated `curryer/kernels/__init__.py` to import and export the new `coverage` module, making it available as part of the public API.


curryer/kernels/coverage.py
```diff
@@ -0,0 +1,319 @@
...
+
+import logging
+import typing
+import warnings
+
+import spiceypy
+from spiceypy.utils.exceptions import SpiceyError
+
+from ..spicierpy import ext
+from ..spicierpy.obj import Body, Frame
```
medley56, line 30: Can you justify why this type of import is necessary? Best practice AFAIK is absolute imports in all cases so it's easy to see where something is coming from.

curryer/kernels/coverage.py
```diff
@@ -0,0 +1,319 @@
...
+
+logger = logging.getLogger(__name__)
+
+# Kernel types that carry time coverage.
+COVERAGE_KERNEL_TYPES = ("SPK", "CK", "PCK")
```
medley56: I mentioned in my review of #169 that I think it's worth creating StrEnum classes for things like this. Because we know that LLMs are predisposed to this kind of implementation, I would also ask for a review of the codebase to find other instances where we are duplicating the same enums or lists of constants and consolidate.

curryer/kernels/coverage.py
```diff
@@ -0,0 +1,319 @@
...
+
+
+class ObjectCoverage(typing.NamedTuple):
+    """Union of coverage windows for one target object.
```
medley56, line 39: "...across all currently loaded kernels"

curryer/kernels/coverage.py
```diff
@@ -0,0 +1,319 @@
...
+
+    Attributes
+    ----------
+    target : int or str or Body or Frame
+        The requested object, as given.
+    windows : tuple of tuple
+        Merged, sorted, disjoint ``(start, stop)`` uGPS windows.
```
medley56, line 46: Is there a particular reason this is coming out as uGPS time and not another system like ET? I'm not sure it matters too much, but I feel like we should try to standardize behavior so that the default time representation across Curryer is consistent.

My personal favorite is TAI seconds, but that's just me.

@BStoneLASP do you have a suggestion on how to maintain consistency here?

curryer/kernels/coverage.py
medley56, line 1: A lot of the functions here need typehinting in their signatures.

medley56 (approved):

curryer/kernels/coverage.py
```diff
@@ -0,0 +1,319 @@
...
+    """Map `target` into a kernel type's ID space, or None when not mappable.
+
+    Mirrors the per-type coercion in :func:`curryer.spicierpy.ext.kernel_coverage`:
+    SPK segments are keyed by body ID, CK by frame ID, and binary PCK by frame
+    *class* ID (integers are used as-is — class IDs have no name lookup — while
+    `Frame`/`Body` targets resolve through the frame info, which requires the
+    frame definitions to be loaded). A target that cannot be resolved is
+    treated as not mappable rather than an error, so scoping stays permissive.
+    """
+    try:
```
medley56, line 124: 🤖 *Comment generated by Claude and reviewed by Gavin*

**Blocking.** `_target_id` catches `SpiceyError`/`ValueError` and returns `None`, and `object_coverage` then just `continue`s — so a target that cannot be resolved at all reports exactly the same thing as a target that is genuinely outside coverage.

Confirmed by running it:

```python
coverage_gaps(['NOT_A_REAL_BODY'], 0, 10**9, kernels=[spk])
# -> ((0, 1000000000),)  + warns "not fully covered"
```

No hint that the body name is bogus. `Body('EARTH')` constructed without an attached frame hits the same path against CK kernels, since `Body.frame` raises `ValueError` when `_frame is None`.

`ObjectCoverage` already carries the signal — `kernels == ()` means nothing contributed. Suggest `valid_window`/`coverage_gaps` collect targets with no contributing kernels and warn (or raise under `error=True`) with a distinct "target not found in any kernel" message.

The same argument applies to the explicit-`kernels` branch at `:81-89`, which silently drops non-DAF files: `pairwise_overlap(text_kernel, spk, target)` returns `()`, indistinguishable from "these two kernels do not overlap."

curryer/kernels/coverage.py
```diff
@@ -0,0 +1,319 @@
...
+        if ktype == "SPK":
+            obj = target.body if isinstance(target, Frame) else target if isinstance(target, Body) else Body(target)
+            return obj.id
+        if ktype == "CK":
+            obj = target.frame if isinstance(target, Body) else target if isinstance(target, Frame) else Frame(target)
+            return obj.id
```
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

`COVERAGE_KERNEL_TYPES` includes `"CK"` and this branch is CK-specific, but no test in `test_kernels_coverage.py` writes a CK — the tests cover SPK and binary PCK only.

CK is the branch with the ID-space subtlety that the PCK branch got an explicit comment and test for: `ext.kernel_objects` returns `ckobj` IDs while `_target_id` supplies `Frame(...).id`, and this also swallows the `Body.frame` `ValueError` noted above. These happen to coincide under curryer's convention (`tests/test_spicierpy/test_spicierpy_ext.py:158-164` queries CK coverage by frame ID/name) — but that's exactly the invariant a test should pin.

Suggest one CK test built the same way as `_write_spk`, covering both an integer target and a `Frame` target.

curryer/kernels/coverage.py
```diff
@@ -0,0 +1,319 @@
...
+        The merged windows and the kernels that contributed. Kernels that do
+        not contain the target contribute nothing.
+
+    """
+    windows = []
+    contributors = []
+    for filename, ktype in _coverage_kernels(kernels):
+        target_id = _target_id(target, ktype)
+        if target_id is None or target_id not in ext.kernel_objects(filename, as_id=True):
+            continue
```
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

Every synthetic kernel in the test file has exactly one segment, so this `zip(flat[0::2], flat[1::2], strict=True)` and the intra-kernel gap merging in `_merge` are never exercised with more than one interval from a single file.

Handling within-kernel gaps via `as_segments=True` is the main thing this module adds over calling `ext.kernel_coverage` directly. I confirmed it works — a two-segment SPK over ET [0,1000]+[2000,3000] yields two disjoint uGPS windows — but it's untested. Suggest one SPK with two disjoint segments, asserting both the two-window union and the resulting gap.

curryer/kernels/coverage.py
```diff
@@ -0,0 +1,319 @@
...
+        Kernel files to consider. Default: the furnished pool.
+
+    Returns
+    -------
+    tuple of tuple
+        Merged, disjoint ``(start, stop)`` uGPS windows; empty when there is
+        no common coverage.
+
+    """
+    windows = None
```
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

`valid_window` calls `object_coverage` once per target, and each call re-runs `_coverage_kernels` (a `getfat` per file, or `ktotal`/`kdata` for the pool) plus a fresh `ext.kernel_objects(filename, as_id=True)` per kernel — a full `spkobj`/`ckobj`/`pckfrm` DAF scan.

For N targets and M kernels that's N x M file scans where M would do. Suggest resolving `_coverage_kernels` once in `valid_window` and threading a `{filename: (ktype, ids)}` map into `object_coverage` (or memoizing per call), which also makes `coverage_rollup` cheaper.

curryer/kernels/coverage.py
```diff
@@ -0,0 +1,319 @@
...
+        Default=False.
+
+    Returns
+    -------
+    tuple of tuple
+        Uncovered ``(start, stop)`` uGPS sub-ranges; empty when the window is
+        fully covered.
+
+    """
+    gaps = _subtract((start, stop), valid_window(targets, kernels=kernels))
```
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

`_subtract` doesn't validate `window`, so `stop <= start` falls straight through to `cursor < stop` being false and returns `()`.

Confirmed: `coverage_gaps([body], 10**9, 0, kernels=[spk])` returns `()` with no warning — a caller that swapped its arguments, or passed a zero-length window, gets a clean "fully covered" answer. Suggest raising `ValueError` in `coverage_gaps` when `stop < start`.

medley56 (changes requested):

Solid module — the interval algebra (`_merge`/`_intersect`/`_subtract`) is correct as far as I could exercise it, it's well documented, and it duplicates nothing: no window arithmetic existed anywhere in `curryer/` before this. All 11 new tests pass locally on the branch.

Requesting changes on the first item only. For a helper whose whole job is making SPICE failures traceable up front, a config typo silently becoming "no coverage" is the wrong default. That's part of a pattern across this stack — see the stack-level note on #177.

**Two things I checked that this code gets right**, so no action needed: text PCKs are reported by `kdata` as ktype `TEXT` (not `PCK`), so the pool branch of `_coverage_kernels` can't feed one into `ext.kernel_objects` and trip its `NotImplementedError`; and kernels that abut exactly in ET do merge into a single window (a shared ET boundary converts to uGPS deterministically, so no spurious 1-microsecond gap).

**Dependency note:** `_target_id` is a deliberate re-implementation of `ext.kernel_coverage`'s per-type target coercion, differing only in returning `None` instead of raising. That's the right call for permissive scoping, but the two ladders must now stay in sync — and #170 is still changing that exact code. Worth a cross-reference comment on the `ext` side, or extracting a shared `ext._coerce_target(target, ktype, strict=...)`.

curryer/kernels/coverage.py
mmaclay: Fair question. Relative is what the rest of the package does, so an absolute import here would be the odd one out — I'd rather this module not be the exception. Happy to convert the whole package as its own change if you'd prefer that as the standard.

curryer/kernels/coverage.py
mmaclay: Done, it uses the new `KernelType` members. The sweep came back mostly clean — `compute.geometry_fields` already does the same thing, and nothing else duplicated kernel types — but the frame-ID logic duplicated between here and `ext.py` was a real instance, and that's one shared helper now.

curryer/kernels/coverage.py
mmaclay: Fixed, and used the same wording everywhere the default means the loaded pool.

curryer/kernels/coverage.py
mmaclay: I picked uGPS to match `ext.kernel_coverage`, which this wraps — it's already curryer's canonical time through `spicetime.adapt`, and introducing a second convention one layer up seemed worse than following the one directly underneath. Keeping it as is, though a repo-wide standard is a fair conversation on its own.

curryer/kernels/coverage.py
mmaclay: Done throughout, including concrete types on the named tuple fields.

curryer/kernels/coverage.py
mmaclay: Agreed — a typo reported as a data gap is the version of this I'd least want in production. A target nothing can resolve is now called out separately from one that's genuinely uncovered, and raises before the gap check under `error=True`. Explicitly-passed files with no time coverage warn too; the pool default still filters quietly, since it legitimately holds text kernels.

curryer/kernels/coverage.py
mmaclay: Fair — that was the one branch resting on an unpinned assumption. There's a CK test over the real attitude kernel now.

curryer/kernels/coverage.py
mmaclay: You're right that this is the main thing the module adds, so it should have been tested from the start. Added a two-segment case covering both the union and the gap.

curryer/kernels/coverage.py
mmaclay: Good spot — kernels resolve once per query now and every target reads from that, which made the rollup cheaper too.

curryer/kernels/coverage.py
mmaclay: Yikes, yes — a swapped-argument caller getting a clean bill of health is the worst possible answer. It raises now.

mmaclay:

The blocking one is fixed: a target nothing can resolve is now reported as exactly that, separately from a target that's genuinely outside coverage, and it raises before the gap check when you ask for errors. A swapped time window raises too, instead of reporting full coverage. Kernels resolve once per query rather than once per target, which also made the rollup cheaper. CK and multi-segment kernels both have tests now, and the signatures are typed.

Two things I left alone: the uGPS default, which matches the layer this wraps — replied on that thread — and the relative import, which is what the rest of the package does. Happy to convert either as its own change.

Ready for another look.

medley56 (approved):

PR #175, opened 2026-08-12
KM PR 4: Add validate_path_length fail-fast check to kernels.path_utils

mmaclay:
This pull request adds a new utility function for validating file path lengths, specifically to catch issues with SPICE kernel paths exceeding their 80-character limit. It also introduces a suite of unit tests to ensure this validation behaves correctly in various scenarios.

**New path length validation utility:**

* Added `validate_path_length` function to `path_utils.py` to provide a fail-fast check that raises a `RuntimeError` if a resolved file path exceeds a configurable maximum length (default 80 characters), with a helpful error message.

**Testing enhancements:**

* Added import for `validate_path_length` in `test_short_paths.py`.
* Added `TestValidatePathLength` test class with comprehensive tests for the new validation function, covering short and long paths, custom limits, and resolution of relative paths.


curryer/kernels/path_utils.py
```diff
@@ -100,6 +100,38 @@ def get_short_temp_dir() -> Path:
...
+    control where their files land.
+
+    Parameters
+    ----------
+    path : Path
+        The path to validate. Made absolute against the working directory
+        before measuring, without dereferencing symlinks — a short symlink
+        (this module's primary shortening strategy) is measured as SPICE
+        sees it, not as its longer target.
+    max_len : int
```
medley56, line 118: Why is this a parameter? Is there a scenario where this would be anything other than 80 chars?

medley56 (approved):

curryer/kernels/path_utils.py
```diff
@@ -100,6 +100,38 @@ def get_short_temp_dir() -> Path:
     return temp_dir
 
 
+def validate_path_length(path: Path, max_len: int = 80) -> None:
```
medley56, line 103: 🤖 *Comment generated by Claude and reviewed by Gavin*

`git grep validate_path_length` across this branch, `km-p1-kernel-cache`, and `km-p1-naif-discovery` returns hits only in `path_utils.py` itself and in `tests/test_kernels/test_short_paths.py`. The docstring's stated audience — "callers that control where their files land" — doesn't exist in the stack.

The obvious candidate is *not* #176's cache: per the top-level comment, cache paths are `furnsh`ed directly and that route tolerates ~255 chars, so validating them against 80 would be a false constraint. The real audience is the meta-kernel writing path — `kernels/writer.py:157` (which already open-codes the same 80-char rule for string values) and `classes.py:_auto_shorten_kernel_paths`.

Suggest wiring it in there so the PR ships a function together with the call site that needs it.

curryer/kernels/path_utils.py
```diff
@@ -100,6 +100,38 @@ def get_short_temp_dir() -> Path:
...
+        Maximum allowed path length (default: 80 for SPICE).
+
+    Raises
+    ------
+    RuntimeError
+        If the absolute path exceeds the maximum length.
+    """
+    path_str = os.path.abspath(path)
+    if len(path_str) > max_len:
+        raise RuntimeError(
```
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

The sibling check for this exact condition — `get_short_temp_dir()` rejecting an over-long `CURRYER_TEMP_DIR` at `:62` — raises `ValueError`, and `ValueError` is the dominant choice repo-wide for bad path/argument input (`kernels/ephemeris.py`, `kernels/classes.py`, `compute/pointing.py`, `spicetime/convert.py`).

A too-long path is invalid input, not an unrelated runtime failure, and as written a caller wanting to catch both length checks needs `except (ValueError, RuntimeError)`. Suggest switching to `ValueError` and updating the `Raises` section.

curryer/kernels/path_utils.py
```diff
@@ -100,6 +100,38 @@ def get_short_temp_dir() -> Path:
...
+            f"Path length ({len(path_str)}) exceeds limit ({max_len}):\n"
+            f"  {path_str}\n"
+            f"Consider setting CURRYER_TEMP_DIR to a shorter base path."
```
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

`CURRYER_TEMP_DIR` only feeds `get_short_temp_dir()`, which is consumed by `update_invalid_paths()` / `copy_to_short_path()` — the path-*rewriting* flow this function explicitly contrasts itself against.

For a caller that "controls where their files land", the advice isn't actionable. Concretely, #176's cache directory derives from `LOCALAPPDATA` / `~/Library/Caches` / `XDG_CACHE_HOME`, none of which `CURRYER_TEMP_DIR` affects. Suggest dropping the hint, or letting the caller supply it (`hint: str | None = None` appended to the message).

curryer/kernels/path_utils.py
```diff
@@ -100,6 +100,38 @@ def get_short_temp_dir() -> Path:
...
+        sees it, not as its longer target.
+    max_len : int
+        Maximum allowed path length (default: 80 for SPICE).
+
+    Raises
+    ------
+    RuntimeError
+        If the absolute path exceeds the maximum length.
+    """
+    path_str = os.path.abspath(path)
```
medley56, line 135: 🤖 *Comment generated by Claude and reviewed by Gavin*

Length semantics diverge from the rest of the module: this measures `os.path.abspath(path)`, while `update_invalid_paths` measures `len(str(item))` on the value as written, and `create_short_symlink`/`copy_to_short_path` compare `len(str(symlink_path))`.

So a relative config path that `update_invalid_paths` passes untouched can fail this check, and the two functions report different lengths for the same input. Absolute measurement is the more defensible choice — worth a `Notes` line calling out the difference now, and a follow-up to make `update_invalid_paths` measure the same way.

medley56 (commented):

Implementation is correct, and measuring `os.path.abspath()` rather than `Path.resolve()` is deliberately right — SPICE opens the symlink path itself, and short symlinks are this module's primary shortening strategy. Tests pass, `ruff check` clean.

## Why 80, when `furnsh` accepts much longer?

This came up in review, so I measured it rather than assuming. There are **three different limits**, and conflating them is the source of the confusion:

| Route | Limit | Failure mode |
|---|---|---|
| `sp.furnsh(path)` directly | ~255 chars | `SpiceFILENAMETOOLONG` — loud |
| Path as a plain string value in a meta-kernel | **80 chars** | **silent** |
| Path split with SPICE's `+` continuation | none observed (153 OK) | — |

The 80 is the SPICE **kernel-pool string-value limit** — which this repo already encodes at `kernels/writer.py:157` (*"Values - Max string value length is 80"*).

What makes it worth defending is the failure mode. Measured:

```
meta-kernel value len= 76 -> furnsh OK, LSK LOADED   (ktotal=2)
meta-kernel value len=120 -> furnsh OK, LSK *MISSING* (ktotal=1)   <-- silent
meta-kernel value len=186 -> furnsh OK, LSK *MISSING* (ktotal=1)   <-- silent
```

The pool truncates the value to 80 characters, `furnsh` on the meta-kernel returns **without error**, and the referenced kernel is simply never loaded. No exception, no warning — the run continues with a missing kernel. That's the strongest justification for keeping this machinery, and it's exactly what `_auto_shorten_kernel_paths` (`kernels/classes.py:257`) exists to prevent.

**Two consequences for this PR:**

1. `validate_path_length` belongs on the **meta-kernel writing path** — `kernels/writer.py` and `classes.py:_auto_shorten_kernel_paths` — not on arbitrary paths. Applied to a path that's only ever `furnsh`ed directly (e.g. #176's cache entries) it's a false constraint, since that route tolerates ~255.
2. **`+` continuation is a real escape hatch and is currently unused.** Verified: a 153-char path split across `'<first half>+', '<second half>'` loads correctly. If long paths keep recurring, teaching `writer.py` to emit continuation syntax removes the limit at its source. Worth a follow-up issue either way.

## Stack note

Despite the "PR 4" label, this branches off `main`, not off #172 — `git merge-base origin/main <branch>` is `6b77600`, which is `main` HEAD. It roots an independent chain and blocks nothing, so it can merge on its own schedule. Neither #176 nor #177 imports anything from here.

curryer/kernels/path_utils.py
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

**Correcting myself on the above.** I said `writer.py:157` "already open-codes the same 80-char rule" and suggested wiring this helper in there. That undersold it — `writer.py` doesn't just encode the rule, it **enforces** it:

```python
# curryer/kernels/writer.py:91-96
issues = []
if not validate:
    logger.warning("Skipping validation of setup file. NOT RECOMMENDED!")
elif not validate_text_kernel(StringIO(setup_txt), issues=issues):
    msg = "\n\t".join(f"{msg}:\n\t\t{ln}" for msg, ln in issues)
    raise ValueError(f"Final text failed validation. Issues[{len(issues)}]:\n\t{msg}")
```

`validate_text_kernel` flags *every* string value over 80 characters (`:159-161`) and this raises `ValueError` before the kernel is written, unless the caller passes `validate=False`.

**Two things follow:**

1. **Curryer-generated setup/meta-kernels are already protected** against the silent truncation I demonstrated, and at a better layer than a per-path pre-check — post-render, covering all string values rather than just paths. My measured silent failure used a **hand-written** meta-kernel, so the real exposure is externally-authored meta-kernels and mission config, not curryer's own output. I should have been clearer about that distinction the first time.

2. **That makes `validate_path_length` closer to redundant than merely uncalled.** The capability isn't missing from the codebase. What a fail-fast helper still buys you is a better error *earlier* — naming the offending path before rendering, rather than surfacing it in a validation dump afterwards — which is a genuine convenience but not a gap. Worth deciding on that basis rather than on "SPICE will break without it".

So please read my recommendation above as "this is where it would go **if** you want the earlier error", not "this check is missing there".

**Related follow-up, since it's the same subsystem:** SPICE's `+` continuation removes the 80-char limit at its source, and curryer doesn't currently emit it. Verified — a 153-char path split across `'<first half>+', '<second half>'` furnishes correctly. If long paths keep recurring, teaching the templates in `writer.py` to emit continuation syntax would make the shortening/symlink machinery unnecessary for the meta-kernel case entirely. Probably its own PR, not this one.

curryer/kernels/path_utils.py
mmaclay: There are actually two limits: a path inside a meta-kernel truncates at 80, but one handed straight to `furnsh` is fine to about 255. Hardcoding 80 would be a false constraint on the direct-furnish callers, so that's in the docstring now rather than only in my head.

curryer/kernels/path_utils.py
mmaclay: Thanks for the follow-up — that changed my answer. Since `writer.py` already enforces the rule after rendering, wiring this in there would be duplication at a worse layer, so I left it alone; the consumer is the kernel manager's static tier in the next stack, which wants the named error before it furnishes anything. The `+` continuation find is great, and I'm filing it separately.

curryer/kernels/path_utils.py
mmaclay: Agreed, especially the inconsistency with the sibling check just above it. It's `ValueError` now.

curryer/kernels/path_utils.py
mmaclay: Right, that hint pointed at machinery this function isn't part of — dropped it. Went with removal over a caller-supplied hint, since no caller has better advice to offer yet.

curryer/kernels/path_utils.py
mmaclay: Good catch, and I agree the absolute measurement is the one to keep. Documented the difference here; making `update_invalid_paths` measure the same way is on the follow-up list, since it changes that function's behavior.

mmaclay:
Thanks for the review, and especially for the follow-up correcting yourself — that changed what I did here.

Since `writer.py` already enforces the 80-character rule after rendering, I've left it alone rather than adding a second check at a worse layer. The helper stays for its earlier, named error; its consumer is the kernel manager's static tier in the next stack. The rest is in: `ValueError` instead of `RuntimeError`, the unactionable env-var hint dropped, and a note explaining why `max_len` is a parameter — the direct-`furnsh` route tolerates far more than 80.

Two follow-ups from your comments: aligning `update_invalid_paths` to measure the same way, and teaching the templates to emit SPICE's `+` continuation.

Ready for another look.

PR #176, opened 2026-08-12
KM PR 5 - add kernel cache

mmaclay:
This pull request adds a robust kernel file caching system to the `curryer.kernels` package, enabling efficient and reliable retrieval of kernel files from local paths, HTTP(S) URLs, and S3 URIs. The cache is versioned per installed package, supports atomic writes, handles network failures gracefully, and can be cleared programmatically. Comprehensive unit tests are included to ensure correct behavior across various scenarios.

**Kernel cache implementation:**

* Added a new `cache.py` module implementing a generic kernel file cache, with support for local, HTTP(S), and S3 sources, per-version cache directories, atomic writes, per-entry max age, and fallback to stale copies on network failure. The cache emits debug logs and only warns when falling back to stale data. (`curryer/kernels/cache.py`)
* Updated the `__init__.py` in `curryer.kernels` to expose the new `cache` module in the package namespace and `__all__`. (`curryer/kernels/__init__.py`)

**Testing and validation:**

* Added a comprehensive unit test suite for the kernel cache, covering local, HTTP, and S3 sources, cache hits/misses, entry aging, network failures, atomicity, and cache clearing. (`tests/test_kernels/test_kernels_cache.py`)


curryer/kernels/cache.py
```diff
@@ -0,0 +1,258 @@
...
+    ----------
+    cache_dir : pathlib.Path, optional
+        Directory to empty. Default: :func:`get_local_cache_dir`.
+
+    Returns
+    -------
+    list of pathlib.Path
+        The removed files.
+    """
+    cache_dir = get_local_cache_dir() if cache_dir is None else Path(cache_dir)
```
medley56, line 121: I would suggest doing some kind of validation to make sure that the directory passed isn't something like `/Users/myuser/home`. Maybe only accept a specific list of possible cache directories? This just blindly unlinks every file in whatever directory you pass it and I don't want that in our public API.

curryer/kernels/cache.py
```diff
@@ -0,0 +1,258 @@
...
+                raise
+            logger.debug("Download attempt %d/%d failed (%s); retrying: %s", attempt, HTTP_ATTEMPTS, error, url)
+            dest.seek(0)
+            dest.truncate()
+            time.sleep(1)
+
+
+def _download_s3(uri: str, dest) -> None:
+    """Stream an S3 object into an open binary file."""
+    import boto3
```
medley56, line 277: Inline import. If Claude is doing this, I'd do a repo-wide sweep to see if it did it elsewhere and we missed it. This leads to extremely confusing code.

curryer/kernels/cache.py
```diff
@@ -0,0 +1,258 @@
...
+# The distribution registers under a different name than the import package.
+DISTRIBUTION_NAME = "lasp-curryer"
+
+
+def _package_version() -> str:
+    """Installed version of this package, whichever name it registered under."""
+    try:
+        return version(DISTRIBUTION_NAME)
+    except PackageNotFoundError:
+        return version(__name__.split(".", 1)[0])
```
medley56: I don't think this is the right place for this functionality. I suggest finding a common place for this to live outside of the caching module.

curryer/kernels/cache.py
```diff
@@ -0,0 +1,258 @@
...
+
+    if entry.is_file():
+        age_sec = time.time() - entry.stat().st_mtime
+        if max_age is None or age_sec < max_age.total_seconds():
+            logger.debug("Cache hit (age %.0fs): %s", age_sec, entry)
+            return entry
+        try:
+            source_size = _source_size(source_str)
+        except Exception as error:
+            logger.warning("Source %s unreachable (%s); using stale cached copy: %s", source_str, error, entry)
```
medley56: I actually think this should rise to the level of a user warning. If users don't have logging configured (which, for users importing a library, is most of the time), they will never see this and it is pretty important if they are relying on updating to a latest kernel reliably.

Also, if users want, it's easy to set `logging.captureWarnings` so user warnings come out in logs.

curryer/kernels/cache.py
```diff
@@ -0,0 +1,258 @@
...
+            entry.touch()
+            logger.debug("Cache hit after size revalidation (%d bytes): %s", source_size, entry)
+            return entry
+        logger.debug("Cache entry stale (source size %s != cached %d): %s", source_size, entry.stat().st_size, entry)
+
+    try:
+        _materialize(source_str, entry)
+    except Exception as error:
+        if entry.is_file():
+            logger.warning("Failed to update %s from %s (%s); using stale cached copy", entry, source_str, error)
```
medley56: Same here. Suggest real warning, not a log message.

curryer/kernels/cache.py
```diff
@@ -0,0 +1,258 @@
...
+                for chunk in resp.iter_content(chunk_size=8192):
+                    dest.write(chunk)
+            return
+        except requests.exceptions.RequestException as error:
+            if attempt == HTTP_ATTEMPTS:
+                raise
+            logger.debug("Download attempt %d/%d failed (%s); retrying: %s", attempt, HTTP_ATTEMPTS, error, url)
+            dest.seek(0)
+            dest.truncate()
+            time.sleep(1)
```
medley56: I think best practice is a back-off retry like 1s, 5s, 20s just in case a server is overloaded, but it's a tiny detail.

tests/test_kernels/test_kernels_cache.py
```diff
@@ -0,0 +1,248 @@
...
+                "get_object",
+                {"Body": StreamingBody(io.BytesIO(payload), len(payload)), "ContentLength": len(payload)},
+                params,
+            )
+        stubber.activate()
+        self.addCleanup(stubber.deactivate)
+        patcher = patch("boto3.client", return_value=client)
+        patcher.start()
+        self.addCleanup(patcher.stop)
+        return stubber
```
medley56: I guess this works, but...

- This Stubber to my eye was designed more for stubbing out temporarily-non-working S3 requests, not for testing. Yes, I know the docs page actually says "for testing" but the semantics of "stubber" does not sound like testing to me; it sounds like stubbing out functionality that is intended to hit a real S3 endpoint someday.
- If this gets messed up at any point, it will make real S3 requests. Granted, they would probably fail because it's not a real bucket, but it would still scare my pants off to learn that a test suite was making real boto calls to S3 buckets with whatever AWS credentials I might have active.
- We can use moto to actually mock out S3, which should prevent calls from actually hitting AWS no matter what.

tests/test_kernels/test_kernels_cache.py
```diff
@@ -0,0 +1,248 @@
...
+        with self.assertNoLogs(cache.logger, level="WARNING"):
+            with self.assertLogs(cache.logger, level="DEBUG") as miss_logs:
+                cache.fetch(HTTP_URL, cache_dir=self.cache_dir)
+            with self.assertLogs(cache.logger, level="DEBUG") as hit_logs:
+                cache.fetch(HTTP_URL, cache_dir=self.cache_dir)
+
+        self.assertTrue(any("Cached" in message for message in miss_logs.output))
+        self.assertTrue(any("Cache hit" in message for message in hit_logs.output))
+
+    def test_clear_cache(self):
```
medley56, line 242: What happens if you accidentally give this an important non-cache directory?

medley56 (changes requested):
I do want to see a couple changes on this one. The blind unlinking of all files in the cache dir that is passed by a user scares me.

curryer/kernels/cache.py
```diff
@@ -0,0 +1,258 @@
...
+    FileNotFoundError
+        If a local source does not exist and there is no cached copy.
+    requests.exceptions.RequestException
+        If an HTTP download fails after retries and there is no cached copy.
+    botocore.exceptions.BotoCoreError or botocore.exceptions.ClientError
+        If an S3 download fails and there is no cached copy.
+    """
+    source_str = str(source)
+    cache_dir = get_local_cache_dir() if cache_dir is None else Path(cache_dir)
+    entry = cache_dir / _basename(source_str)
```
medley56, line 187: 🤖 *Comment generated by Claude and reviewed by Gavin*

Not reopening the resolved cache-key thread — just confirming the residual risk is understood, since commit `2c3381f` clarified the docstring but left the key as `basename`.

Concretely: `tests/test_kernels/test_kernels_cache.py:130` (`test_warm_hit_no_network`) has a fetch of an **HTTP URL** return the bytes of an unrelated **local** file that happened to share a basename, with no network call. With the default `max_age=None` a warm entry is returned with no validation at all — first writer wins permanently.

The docstring's justification (NAIF generic kernels version their filenames) holds for NAIF, but not for local paths or mission S3 buckets, where `naif0012.tls`, `*_meta_v01.tm`, and per-mission SPKs collide routinely — and the failure mode is silently loading the wrong kernel into SPICE, not an error.

If NAIF-only is the intended scope, that's a reasonable call — worth stating in the docstring so the next reader doesn't re-raise it. If not, `cache_dir / sha256(source)[:12] / basename` keeps the extension SPICE needs while making distinct sources distinct.

curryer/kernels/cache.py
```diff
@@ -0,0 +1,258 @@
...
+        except Exception as error:
+            logger.warning("Source %s unreachable (%s); using stale cached copy: %s", source_str, error, entry)
+            return entry
+        if source_size is not None and source_size == entry.stat().st_size:
+            entry.touch()
+            logger.debug("Cache hit after size revalidation (%d bytes): %s", source_size, entry)
+            return entry
+        logger.debug("Cache entry stale (source size %s != cached %d): %s", source_size, entry.stat().st_size, entry)
+
+    try:
```
medley56, line 205: 🤖 *Comment generated by Claude and reviewed by Gavin*

`_materialize` `os.replace`s new content onto the *same* path an earlier `fetch()` already handed to a caller. If that path is currently furnished, SPICE keeps serving the previously parsed pool contents (text kernels) or its open handle on the now-unlinked inode (binary kernels) — so a "successful" refresh is invisible to the running process.

This is related to the `clear_cache` scoping concern raised elsewhere, but distinct: it bites even with a perfectly-scoped cache directory, because the problem is in-place replacement under SPICE's process-global pool rather than deletion.

The module docstring advertises use "through the kernel manager", but nothing here talks to `spicierpy.ext.KernelSet`. Suggest at minimum documenting that callers must unload-and-refurnish after a `fetch()` that re-downloaded; better, return a flag (or a per-content path) so callers can distinguish a refresh from a hit.

curryer/kernels/cache.py
```diff
@@ -0,0 +1,258 @@
...
+        age_sec = time.time() - entry.stat().st_mtime
+        if max_age is None or age_sec < max_age.total_seconds():
+            logger.debug("Cache hit (age %.0fs): %s", age_sec, entry)
+            return entry
+        try:
+            source_size = _source_size(source_str)
+        except Exception as error:
+            logger.warning("Source %s unreachable (%s); using stale cached copy: %s", source_str, error, entry)
+            return entry
+        if source_size is not None and source_size == entry.stat().st_size:
```
medley56, line 199: 🤖 *Comment generated by Claude and reviewed by Gavin*

Byte size is a weak validator: a same-size regeneration upstream is accepted as current, *and* the `entry.touch()` on the next line resets the clock — so a stale copy stays warm indefinitely.

For local sources, `_source_size` (`:183`) already stats the file but discards the free `st_mtime` comparison. For HTTP, `Content-Length` is unreliable under `Content-Encoding: gzip` (compressed size vs. stored size, so perpetual re-download) and absent under chunked transfer.

This is the condition the suggested user-warning at `:146` would be reporting on, so it's worth improving the detection alongside it: compare `(size, mtime)` for local sources, and use `ETag`/`Last-Modified` with a conditional GET (`If-None-Match`/`If-Modified-Since` -> 304) for HTTP, falling back to the size check.

medley56 (commented):

The plumbing here is good — atomic temp-file + rename, retries, and the offline stale-copy fallback are well implemented and genuinely well tested. The existing open comments already cover the API-safety surface (`clear_cache` scope, warnings vs. logs, backoff, moto vs. Stubber, the inline import), so I haven't repeated any of those. What's below is only what they don't cover.

**One correction to flag:** an earlier draft of this review called out missing 80-char path validation as a major issue. That was wrong. Cache paths are handed to `furnsh` directly, and that route tolerates ~255 chars — the 80-char limit applies to meta-kernel *string values*, which this module never emits. Measurements are in the top-level comment on #175. **No path validation is needed here.**

The useful consequence: **#175 isn't actually a dependency of this PR.** `cache.py` imports nothing from `path_utils`, and correctly so. Chain B is stacked by convenience rather than need, so #176/#177 could rebase off `main` directly if that's ever helpful.

**Merge-conflict heads-up:** this chain and #172 both edit the same two lines of `curryer/kernels/__init__.py` (the import and `__all__`). Whichever lands second will conflict; the union is `attitude, cache, classes, coverage, create, discovery, ephemeris, writer`.

curryer/kernels/cache.py
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

Agreeing with the placement comment on this function, plus a bug: **this fallback cannot work.** The import package is `curryer`, but the only registered distribution is `lasp-curryer`, so `version("curryer")` raises `PackageNotFoundError` too. Verified locally:

```
curryer      -> PackageNotFoundError
lasp-curryer -> 0.5.1
```

So in a non-installed source checkout, `get_local_cache_dir()` raises instead of falling back, and no test covers the branch. This becomes load-bearing in #177, which puts `get_local_cache_dir()` on `find_default_file()`'s path — the function reached from every `spicetime.adapt()`.

There's also a second reason to move it, beyond tidiness: `curryer/kernels/__init__.py` eagerly imports `attitude`/`classes`/`create`/`ephemeris`, each doing `from .. import spicetime` at module scope. So any import of `kernels.cache` cycles back through `spicetime` — which is exactly why #177 has to defer its imports. `cache.py` itself imports nothing from `spicetime`; the cycle is purely an artifact of the eager package `__init__`.

Moving `_package_version` (and ideally the cache-dir resolution) into a neutral, dependency-free module that `kernels/__init__` doesn't pull in fixes the placement complaint, gives the fallback a real default, and lets #177 drop both deferred imports. Suggest one follow-up covering all three.

curryer/kernels/cache.py
```diff
@@ -0,0 +1,258 @@
...
+
+    Returns
+    -------
+    list of pathlib.Path
+        The removed files.
+    """
+    cache_dir = get_local_cache_dir() if cache_dir is None else Path(cache_dir)
+    removed = []
+    if cache_dir.is_dir():
+        for cached_file in cache_dir.iterdir():
```
medley56, line 129: 🤖 *Comment generated by Claude and reviewed by Gavin*

Beyond the directory-scoping concern already raised on this function, there's a growth problem: the cache has no size cap, no LRU, and no eviction. Because the directory is keyed by installed version, **every release starts a fresh dir and permanently orphans the previous one** — with DE-series SPKs at 100+MB, that's real disk.

`clear_cache()` only unlinks regular files in the single directory it's given (default: the current version's), so it never reclaims sibling version dirs and leaves subdirectories behind.

A `prune_old_versions()` — or having `clear_cache()` prune siblings under the package cache root by default — would pair well with the allow-list validation suggested above, since both are about making this function's blast radius explicit.

medley56:

## Follow-up: one refactor closes three open threads

Raising this here rather than as a separate issue, since `cache.py` is where the knot is. Three
things flagged across this PR and #177 have a **single root cause**:

1. `_package_version` living in the caching module (raised on `cache.py:50` here)
2. Its `PackageNotFoundError` fallback being unreachable — it retries `version("curryer")`, but the
   distribution is `lasp-curryer`, so that raises too
3. The two deferred imports in #177 (`leapsecond.py:60` and `:182`), which prompted "why is this
   import deferred?" on both

**The root cause:** `curryer/kernels/__init__.py` eagerly imports `attitude`, `classes`, `create`,
`ephemeris`, and every one of those does `from .. import spicetime` at module scope. So *any*
import of `curryer.kernels.cache` pulls in the whole `kernels` package and cycles back through
`spicetime` — which is why `leapsecond` has no choice but to defer.

Worth stressing: **`cache.py` imports nothing from `spicetime`.** There is no real dependency
between them. The cycle is purely an artifact of the eager package `__init__`, so the deferrals
are working around a structural quirk rather than a genuine layering constraint.

**Suggested shape:** move the dependency-free helpers — version lookup and cache-dir resolution —
into a neutral low-level module that `kernels/__init__` does not import (something like
`curryer/_paths.py` or `curryer/utils/`). That:

- puts `_package_version` somewhere defensible
- lets the fallback become a real default (e.g. `"unknown"`) with a test covering the branch
- lets #177 drop **both** deferred imports and import normally at module scope

One change, three threads closed. Probably worth doing before #177 merges, so the deferred imports
never land — but not blocking on it if you'd rather keep the stack moving and clean up after.

curryer/kernels/cache.py
mmaclay: You're right, that doesn't belong in a public API as written. It refuses anything outside the curryer cache root now, with a test that hands it an unrelated directory and checks the files survive.

curryer/kernels/cache.py
mmaclay: Did the sweep, and the answer surprised me: inline boto3 is the existing pattern here — `correction/io.py` and `correction/dataio.py` both predate this work. Left it consistent with them rather than making this one file different, but happy to hoist all of them up as a separate pass if you'd rather that be the standard.

curryer/kernels/cache.py
mmaclay: Agreed — that and the cache-dir resolution live in a new `curryer/utils.py` now, which imports nothing internal. The same move fixed the broken fallback and the import cycle you found below.

curryer/kernels/cache.py
mmaclay: Agreed, and the point about library users without logging configured is the right frame — quietly serving a stale leapsecond kernel is exactly when someone needs to be told. Both fallbacks warn properly now.

curryer/kernels/cache.py
mmaclay: Same — both warn now, and the tests check for the warning rather than a log record.

curryer/kernels/cache.py
mmaclay: Not a tiny detail against a public NAIF endpoint — took your 1/5/20 exactly, with a test on the sleeps so it can't quietly regress.

tests/test_kernels/test_kernels_cache.py
mmaclay: Your second point is the persuasive one; not worth carrying that risk for convenience. It's moto now, and a nice side effect is that the same-size-different-bytes case became a real test instead of proving itself by an absent stub.

tests/test_kernels/test_kernels_cache.py
mmaclay: As of this commit, an error before anything gets unlinked — with a test for exactly that. Before it, what you're imagining.

curryer/kernels/cache.py
mmaclay: Thanks for closing the loop rather than reopening it, and your read of that warm-hit test is fair. NAIF-only is the intended scope, so the module docstring says so now along with the escape hatch; hashed subdirectories are the right answer if we ever point this at mission buckets.

curryer/kernels/cache.py
mmaclay: Subtle one, hadn't thought about it. Documented for now — a refresh swaps the file underneath anyone already holding it, so callers need to unload and refurnish. I stopped short of the return flag because the kernel manager in the next stack owns furnish state, but happy to add it now if you'd rather.

curryer/kernels/cache.py
mmaclay: Agreed on all of it, especially gzip causing perpetual re-downloads. Filing it as a follow-up rather than folding it in here — say the word if you'd rather it block this one.

curryer/kernels/cache.py
mmaclay: Most useful comment on this PR — you found a bug and the reason I wrote it. Both functions moved to a dependency-free `curryer/utils.py`, so the fallback is real, it can't raise, and #177 drops its deferred import entirely.

curryer/kernels/cache.py
mmaclay: Good point about every release orphaning the last one — that adds up fast with DE kernels. Follow-up issue; the scoping fix landing here is the prerequisite anyway, since it makes clearing the whole root an explicit thing to ask for.

mmaclay:
Thanks for the review — pushed changes for all of it, and your proposed refactor does close all three threads.

`package_version` and the cache-dir resolution now live in a new dependency-free `curryer/utils.py`. That gives them a defensible home, makes the fallback real so it can't raise, and lets #177 import normally instead of deferring.

`clear_cache` refuses anything outside the curryer cache root, both stale fallbacks warn rather than log, retries back off 1/5/20, and the S3 tests use moto so they can't reach real AWS under whatever credentials happen to be active.

Two follow-ups filed rather than folded in: proper revalidation with a conditional GET, and pruning orphaned version directories. The inline boto3 I've left, since it matches `correction/` — replied on that thread.

Ready for another look.

medley56 (approved):
Assuming the test pass, LGTM

mmaclay:
One Note from MW: there is a strange logging error / output issue in the Libera Utils version of KM caching. We should run this and test for clear and consistent logging output, and ensure our cache hits / misses are expected and explainable.

PR #177, opened 2026-08-12
KM PR 6 Add NAIF generic-kernel discovery and Leapsecond caching

mmaclay:
This pull request introduces a new mechanism for discovering and updating NAIF generic kernel files (such as leapsecond kernels) by scraping the NAIF server or a mirror for the latest versions. The update refactors leapsecond kernel management to use a local cache and the new discovery utilities, improving reliability and maintainability. The most important changes are summarized below.

**New NAIF kernel discovery utilities:**

* Added `curryer/kernels/discovery.py`, which provides functions to locate the most recent NAIF kernel files by scraping index pages, supporting versioned filenames and subdirectory routing.
* Updated `curryer/kernels/__init__.py` to include the new `discovery` module in the package exports.

**Refactoring leapsecond kernel management:**

* Refactored `curryer/spicetime/leapsecond.py` to use the new discovery and cache utilities:  
  - The default leapsecond kernel is now found by searching both the packaged data directory and the local kernel cache, ensuring the most recent file is used. [[1]](diffhunk://#diff-5fdecf7044559283168ec8b31e8c85984e54f9a8fea71e37793793b09ecb0ec9L7-R8) [[2]](diffhunk://#diff-5fdecf7044559283168ec8b31e8c85984e54f9a8fea71e37793793b09ecb0ec9R39-R44) [[3]](diffhunk://#diff-5fdecf7044559283168ec8b31e8c85984e54f9a8fea71e37793793b09ecb0ec9L49-R74)
  - The update process downloads new kernels into the local cache, and the cache is checked before downloading.
  - Kernel update checks now use the new discovery utilities to find the latest version, rather than scraping manually.

**Data and configuration updates:**

* Added the latest NAIF leapsecond kernel file (`naif0012.tls`) to the data directory.
* Updated internal paths and documentation to reflect the new cache-based approach for kernel management.

These changes improve the robustness and updatability of kernel management, and centralize NAIF kernel discovery logic for easier reuse and extension.


curryer/data/generic/naif0012.tls
medley56, line 1: Is this here as a fallback just because it's used so frequently and isn't very big? I thought we already had the LSK. Is this adding 0012 alongside 0011? If so, we should drop 0011.

curryer/spicetime/leapsecond.py
```diff
@@ -155,31 +166,25 @@ def update_file():
...
         return None
 
-    # Form the URL to download and the destination filename.
-    kernel_url = LEAPSECOND_BASE_URL + kernel_name
-    default_file = find_default_file()
-    kernel_file = default_file.with_name(kernel_name)
-    if kernel_file.is_file():
-        raise FileExistsError(f"New file already exists, but wasn't the default! File: {kernel_file}")
+    # Deferred import: curryer.kernels imports spicetime at package init.
+    from ..kernels import cache as kernels_cache
```
medley56, line 195: Why do we need this import deferred?

curryer/spicetime/leapsecond.py
```diff
@@ -46,17 +52,26 @@ def find_default_file():
...
     elif os.getenv("LEAPSECOND_FILE_ENV", None):
-        path = Path(os.getenv("LEAPSECOND_FILE_ENV"))
+        search_dirs = [Path(os.getenv("LEAPSECOND_FILE_ENV"))]
     else:
-        path = Path(__file__).parent
-        path = path.joinpath(_LEAPSECOND_FILE_PATH).resolve()
-    leapsecond_files = list(path.glob(_LEAPSECOND_FILE_GLOB))
-    leapsecond_files.sort()
+        # Deferred import: curryer.kernels imports spicetime at package init.
+        from ..kernels import cache as kernels_cache
```
medley56: Why is this import deferred?

curryer/spicetime/leapsecond.py
```diff
@@ -155,31 +166,25 @@ def update_file():
...
-    # Download the latest leapsecond file.
+    # Download the latest leapsecond file into the local kernel cache.
+    kernel_url = LEAPSECOND_BASE_URL + kernel_name
     logger.debug("Downloading kernel file: %r", kernel_url)
-    resp = requests.get(kernel_url, timeout=60)
-    resp.raise_for_status()
-
-    logger.debug("Saving kernel file: %r", kernel_file)
-    kernel_file.write_text(resp.text)
+    kernel_file = kernels_cache.fetch(kernel_url)
```
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

**This is the one I'd hold on.** `cache.get_local_cache_dir()` is keyed by installed package version (`~/.cache/curryer/<version>/`), so the kernel this downloads is only visible to `find_default_file()` until curryer itself is upgraded — after which the search silently falls back to the packaged `naif0012.tls`.

That's fine for a rolling binary PCK, but wrong for leapseconds. Once NAIF publishes `naif0013.tls`, a user who ran `update_file()` gets correct conversions until they `pip install -U lasp-curryer`, at which point they **silently regress to a kernel missing a leapsecond**. No error, just wrong times — the same failure class as the rest of this stack.

The docstring at `:170-172` ("`find_default_file` will resolve to it going forward") overstates this.

Suggest storing the LSK in a version-independent location (e.g. `cache_dir.parent / "lsk"`, or a `cache.fetch(..., cache_dir=...)` override), or at minimum documenting the expiry and logging a warning when the packaged kernel wins over a previously-updated one.

curryer/spicetime/leapsecond.py
```diff
@@ -46,17 +52,26 @@ def find_default_file():
...
     else:
-        path = Path(__file__).parent
-        path = path.joinpath(_LEAPSECOND_FILE_PATH).resolve()
-    leapsecond_files = list(path.glob(_LEAPSECOND_FILE_GLOB))
-    leapsecond_files.sort()
+        # Deferred import: curryer.kernels imports spicetime at package init.
+        from ..kernels import cache as kernels_cache
+
+        search_dirs = [Path(__file__).parent.joinpath(_LEAPSECOND_FILE_PATH).resolve()]
+        cache_dir = kernels_cache.get_local_cache_dir()
```
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

This puts `cache._package_version()` on the path of the library's most load-bearing function — `find_default_file()` is reached from every `spicetime.adapt()` via `leapsecond.load()`, and from `kernels/classes.py:82`.

Combined with the `PackageNotFoundError` bug flagged on #176 (the fallback calls `version("curryer")`, but the distribution is `lasp-curryer`, so it raises too), a source tree used via `PYTHONPATH` with no `pip install` now propagates that exception straight out of `find_default_file()`.

CI (`pip install ".[dev,test]"`) and the Dockerfile both install, so this isn't a live break — but it's a new failure mode for a function that previously needed nothing but the filesystem. Fixing `_package_version` in #176 covers it; a `try/except` around these lines (log-and-skip the cache dir) would too.

curryer/kernels/discovery.py
```diff
@@ -0,0 +1,170 @@
...
+# Index subdirectory for each kernel file extension, relative to the base URL.
+NAIF_SUBDIR_BY_EXTENSION = {
+    "tls": "lsk/",
+    "tpc": "pck/",
+    "bpc": "pck/",
+    "tf": "fk/planets/",
+    "bsp": "spk/planets/",
+}
+
+HTTP_ATTEMPTS = 3
```
medley56, line 41: 🤖 *Comment generated by Claude and reviewed by Gavin*

`HTTP_ATTEMPTS = 3` / `HTTP_TIMEOUT_SEC = 30` and the try/`raise_for_status`/`time.sleep(1)` loop at `:118-128` are byte-for-byte the policy in the sibling module `cache._download_http`, added one PR earlier in this same chain.

Concretely: the exponential-backoff suggestion on #176 would need to be applied here as well to take effect, and the two copies will drift the first time someone tunes a timeout.

Suggest hoisting the constants and a small `_get_with_retries` helper into one module (or `kernels/__init__`) and importing it in the other.

curryer/spicetime/leapsecond.py
```diff
@@ -131,22 +146,18 @@ def check_for_update():
...
+    # Deferred import: curryer.kernels imports spicetime at package init.
+    from ..kernels import discovery
 
-    files = re.findall(r'href="(naif[0-9]{4}\.tls)"', resp.text)
-    if len(files) == 0:
-        raise ValueError(f"No files were found on the NAIF page: {LEAPSECOND_BASE_URL!r}")
-
-    files.sort()
-    logger.debug("Found files on NAIF page: %r", files)
+    latest_url = discovery.find_most_recent_naif_kernel(LEAPSECOND_BASE_URL, discovery.NAIF_LSK_REGEX)
```
medley56: 🤖 *Comment generated by Claude and reviewed by Gavin*

`check_for_update()` gets a fully-resolved `latest_url` from discovery, strips it back to a basename, and then `update_file()` re-derives the URL as `LEAPSECOND_BASE_URL + kernel_name` at `:185`.

The round trip is lossy — it only works because `LEAPSECOND_BASE_URL` happens to end in `/` — and it means the module's subdirectory routing (`naif_index_url()` and `find_latest_naif_kernel_url()`, the two entry points the docstring advertises) is exercised only by tests.

Suggest having `update_file()` call `discovery.find_latest_naif_kernel_url(discovery.NAIF_LSK_REGEX)` and hand that URL straight to `cache.fetch`, keeping `check_for_update()`'s string return for API compatibility.

curryer/data/generic/naif0012.tls
```diff
@@ -0,0 +1,152 @@
+KPL/LSK
```
medley56, line 1: 🤖 *Comment generated by Claude and reviewed by Gavin*

Answering the question raised on this file: there is no `naif0011.tls` anywhere in the repo — `git ls-tree` over `main` shows a single LSK, `data/generic/naif0012.tls`. So this isn't adding 0012 alongside an older kernel; it's a **second, byte-identical copy of the same version**.

Vendoring it *inside* the package is the right fix (the old repo-root path was outside the wheel entirely). The issue is that the root copy still exists and is still referenced by `tests/test_correction/*` (`generic_kernel_dir=Path("data/generic")`), `examples/correction/`, `tests/sit4_l1a_geolocation.py`, the Dockerfile (`COPY data $DATA_DIR`), and `docs/source/users.md:37,50`.

Nothing keeps the two in sync, so a future `update_file`-style refresh of one leaves the other stale, and it isn't obvious to a contributor which is authoritative. Suggest deleting the repo-root copy and pointing the dev-data consumers at `spicetime.leapsecond.find_default_file()`, or at minimum a one-line note in `docs/source/users.md` naming the packaged copy as the source of truth.

curryer/spicetime/leapsecond.py
```diff
@@ -46,17 +52,26 @@ def find_default_file():
     #   Pylint thinks the `Path` class is really a `PurePath`.
     # pylint: disable=no-member
     if LEAPSECOND_USER_FILE_PATH is not None:
-        path = Path(LEAPSECOND_USER_FILE_PATH)
+        search_dirs = [Path(LEAPSECOND_USER_FILE_PATH)]
```
medley56, line 68: 🤖 *Comment generated by Claude and reviewed by Gavin*

Nit: when `LEAPSECOND_FILE_ENV` or `LEAPSECOND_USER_FILE_PATH` is set, the cache dir is deliberately excluded from the search — so `update_file()` downloads a new kernel that `find_default_file()` will never return, and `check_for_update()` then reports the same available update forever.

The precedence itself is defensible (an explicit override should win), but it contradicts the new `update_file()` docstring. Suggest a `logger.warning` (or user warning) in `update_file()` when an override is active.

medley56 (commented):

Taking this one as the capstone, so this covers both the PR itself and the six-PR stack as a whole.

## Answers to the two open questions

**"Is this adding 0012 alongside 0011? If so, we should drop 0011."**
No 0011 exists. `git ls-tree` over `main` shows exactly one LSK in the repo: `data/generic/naif0012.tls`. This PR adds a **second, byte-identical copy of the same version** at `curryer/data/generic/naif0012.tls`. So the thing to drop isn't an older kernel — it's the now-redundant repo-root copy. See the inline comment on the data file.

**"Why are these imports deferred?"** (`leapsecond.py:60` and `:182`)
Genuine circular-import avoidance, and the inline comment is accurate. `curryer/kernels/__init__.py` eagerly imports `attitude`, `classes`, `create`, `ephemeris`, and every one of those does `from .. import spicetime` at module scope — so importing `kernels.cache` from `spicetime.leapsecond` at module level cycles.

Worth noting `cache.py` imports nothing from `spicetime` itself; the cycle is an artifact of the eager package `__init__`, not a real dependency. This is the **same root cause** as the `_package_version` placement comment on #176, and one follow-up fixes both: move the dependency-free helpers into a neutral module that `kernels/__init__` doesn't import. Until then, the deferrals are correct and should stay.

## This PR

Design is sound and the 40 tests pass. The packaging change is a **real fix, not just a move**: `_LEAPSECOND_FILE_PATH` previously pointed at repo-root `data/generic`, which is *outside* the `curryer` package — so published wheels shipped no leapsecond kernel at all. Built the branch with `poetry build` and confirmed `curryer/data/generic/naif0012.tls` is now in the wheel (poetry-core ships non-Python files inside the package dir, so no MANIFEST/`include` entry is needed).

Also checked against the live NAIF index: `naif0012.tls` is still the current LSK, and all five `NAIF_*_REGEX` constants match real filenames in today's `lsk/` and `pck/` listings.

The test rewrite is a net gain — `responses` instead of `MagicMock(spec=Response)`, isolated cache dirs so a warm dev-machine cache can't perturb resolution, and real package/env/cache precedence coverage. The one deleted case (`test_update_file_fails_if_new_file_exists`) covered a `FileExistsError` branch this PR removes, so nothing real is lost.

## The stack, read as one feature

**It's two chains, not one.** Despite the 1-6 labelling, `#175`'s merge-base is `main` HEAD, so it roots an independent chain:

```
main --+-- #169 -- #170 -- #172        (chain A)
       +-- #175 -- #176 -- #177        (chain B)
```

Chain A is a genuine dependency chain — #170 implements the binary-PCK branch #172 calls, and #172 consumes #169's `loaded_kernels`. Chain B mostly isn't: only the #176 -> #177 edge is real. `validate_path_length` is referenced by neither #176 nor #177, and correctly so (details on #175).

**Expect one merge conflict.** #172 and #177 both edit the same two lines of `kernels/__init__.py`. Union: `attitude, cache, classes, coverage, create, discovery, ephemeris, writer`.

**The recurring theme is silent failure.** Across the stack: a misspelled coverage target and an inverted window both report as ordinary answers (#172); the basename cache key can return a different kernel's bytes (#176); and an updated LSK silently reverts on upgrade (finding 1 below). The template for all of these is pre-existing and worth naming — a >80-char path in a meta-kernel truncates in the pool, `furnsh` succeeds, and the kernel never loads (measured; see #175). For a geolocation library these produce wrong numbers with no traceback. That's the class of bug this stack is fighting, and it's the strongest argument for the "make unknown loud" fixes.

**Duplication about to calcify.** The HTTP retry policy here is byte-for-byte `cache._download_http` from the immediately preceding PR — so the exponential-backoff request on #176 would have to be applied twice to take effect. Worth unifying inside the stack.

**Suggested merge order:** #169 -> #170 -> #172 (land #170's `MOON_PA` fix first — it's a real wrong-answer bug on kernels this repo ships), then #176 -> #177, with #175 whenever a caller exists.

**Worth follow-up issues rather than review comments:** `docs/source/changelog.md` has no entry for any of chain B, though it adds two public modules, a packaged data file, and a behaviour change to `update_file()`; `docs/source/users.md:37,50` still describes `data/generic` as the leapsecond kernel's home; and `NAIF_DE_REGEX` / `NAIF_HIGH_PREC_PCK_REGEX` / `NAIF_EARTH_EXTENDED_PCK_REGEX` are defined and tested but have no consumer yet — presumably a follow-on PR, worth saying so in the description so they don't read as dead constants. No licensing concern with vendoring the LSK (NAIF generic kernels are U.S.-government public-domain data), though a line in `NOTICE` crediting NAIF/JPL would be tidy.

medley56:

## Follow-up: docs debt and dangling constants for this chain

Collecting these here rather than as separate issues, since this PR is the tip of the chain that
introduces all of it. None of it blocks the merge — but the first item is the kind of thing that
silently never gets done once the stack lands.

### 1. No changelog entry for any of chain B

`docs/source/changelog.md` has no entry covering #175, #176, or #177, even though between them they
add **two public modules** (`kernels.cache`, `kernels.discovery`), a **packaged data file**, and a
**behaviour change** to `update_file()` — which now downloads into a local cache rather than
alongside the package. That last one is user-visible and worth calling out explicitly, including
the version-keyed expiry caveat noted in the review above.

### 2. `users.md` still describes the old kernel location

`docs/source/users.md:37,50` still describes `data/generic` as where the leapsecond kernel lives.
After this PR the authoritative copy is the packaged `curryer/data/generic/naif0012.tls`, resolved
via `spicetime.leapsecond.find_default_file()`. Since the repo-root copy also still exists (see the
inline comment on the data file), the docs are now the main thing telling a contributor which one
is real — worth making them say the right thing.

### 3. Three discovery constants have no consumer

`NAIF_DE_REGEX`, `NAIF_HIGH_PREC_PCK_REGEX`, and `NAIF_EARTH_EXTENDED_PCK_REGEX` are defined and
tested, but nothing in the codebase uses them — only `NAIF_LSK_REGEX` is wired up, through
`leapsecond`. I verified all five match real filenames in today's NAIF `lsk/` and `pck/` listings,
so they're correct, just unused.

Presumably the point of a follow-on PR that brings DE/PCK kernels under the same
discovery-plus-cache treatment. Worth one line in the PR description saying so, otherwise they read
as dead constants to the next person through — and dead-looking constants tend to get "cleaned up"
by someone who doesn't know the plan.

### 4. Minor: `NOTICE` attribution for the vendored kernel

No licensing concern with vendoring `naif0012.tls` — NAIF generic kernels are U.S.-government
public-domain data. But now that the repo ships a NAIF-produced file inside the package, a line in
`NOTICE` crediting NAIF/JPL would be tidy.

curryer/data/generic/naif0012.tls
mmaclay: No older one to drop — there's only ever been 0012 in the repo. This is a second copy of the same file, vendored inside the package so it actually ships in the wheel. More on the duplication in your other thread.

curryer/spicetime/leapsecond.py
mmaclay: Fair question — it deserved a comment in the code rather than an answer here. The package `__init__` imports modules that import `spicetime`, so this cycled back; your suggestion on the cache PR fixed it at the root, and this one is a normal import now.

curryer/spicetime/leapsecond.py
mmaclay: Same cause as the thread above. This one is still deferred because that cycle is real, but it explains itself in a comment now.

curryer/spicetime/leapsecond.py
mmaclay: You were right to hold this. Silently regressing to a kernel missing a leapsecond after a routine upgrade — no error, just wrong times — is the worst bug in either stack. Updated kernels go somewhere version-independent now, with a test that fakes an upgrade and checks the update still wins, and the overstated docstring is fixed.

curryer/spicetime/leapsecond.py
mmaclay: Fixed at the root on the cache PR — that function can't raise anymore, so the source-tree failure mode is gone and this one no longer defers anything.

curryer/kernels/discovery.py
mmaclay: Correct, and your prediction came true immediately: the backoff would have missed this copy. One shared helper now, used by both.

curryer/spicetime/leapsecond.py
mmaclay: Agreed — "works because the constant happens to end in a slash" is a good description of a latent bug. It resolves the URL once and hands that straight over, so the subdirectory routing gets exercised in real use rather than only by tests.

curryer/data/generic/naif0012.tls
mmaclay: Your read is right and I want to get there, but separately: removing the root copy breaks 37 tests across 7 files, because the fixtures build that path piece by piece so the filename never appears in a search. For now I took your fallback and made the docs name the packaged copy as authoritative. Filing the consolidation so it doesn't evaporate — two unsynchronized copies isn't a resting place.

curryer/spicetime/leapsecond.py
mmaclay: Small but real; reporting the same available update forever is a confusing way to spend an afternoon. It warns now when an override is active, and the precedence is unchanged.

mmaclay:
Thanks for the review — the leapsecond hold was the right call.

Updated leapsecond kernels now cache somewhere version-independent, so a `pip install -U` can't silently drop you back to a kernel missing a leapsecond. There's a test that fakes an upgrade and checks the update still wins. `update_file` resolves the URL once and hands it straight over instead of rebuilding it from a basename, warns when an override would hide the download, and discovery shares one retry helper with the cache instead of keeping its own copy.

From your follow-up: changelog, `users.md`, and the NAIF credit in `NOTICE` are all in. The changelog entry covers both chains but only lives on this branch, so chain A's entry arrives whenever this merges. The unused discovery patterns feed the manager's NAIF tier in the next stack — noted so nobody cleans them up.

Consolidating the duplicate LSK is filed separately; it breaks 37 tests and deserves its own review.

Ready for another look.

medley56 (approved):

